PR #13, opened 2020-01-22
Add get Ethereum network name

davidalbela:
- Add `EthereumNetworkName` enum.
  - Add `get_network_name` method in EthereumClient class.
  - Add `test_get_ethereum_network` test.

### DIFF
--- a/gnosis/eth/ethereum_client.py
+++ b/gnosis/eth/ethereum_client.py
@@ -4,6 +4,7 @@ from typing import Any, Dict, List, NamedTuple, Optional, Union
 
 import eth_abi
 import requests
+from enum import Enum
 from eth_abi.exceptions import InsufficientDataBytes
 from eth_account import Account
 from eth_account.signers.local import LocalAccount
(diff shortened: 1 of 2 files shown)


gnosis/eth/ethereum_client.py
```diff
@@ -634,6 +653,16 @@ def get_slow_provider(self, timeout: int):
         else:
             return self.w3_provider
 
+    def get_network_name(self):
```
Uxio0: I would return the enum instead of the string.

Please add the type info `def get_network_name(self) -> EthereumNetworkName:`

gnosis/eth/ethereum_client.py
```diff
@@ -634,6 +653,16 @@ def get_slow_provider(self, timeout: int):
...
+        """
+        Get network name based on the network version id
+        :return: The name of the current Ethereum network
+        """
+        if isinstance(self.w3, Web3):
```
Uxio0: This is not needed, `self.w3` is always a `Web3` instance

gnosis/eth/ethereum_client.py
```diff
@@ -634,6 +653,16 @@ def get_slow_provider(self, timeout: int):
...
+            return EthereumNetworkName(self.w3.net.version).name
```
Uxio0: `return EthereumNetworkName(self.w3.net.version)`

PR #15, opened 2020-01-28
Adjust basegas price after hardfork

Uxio0:
- Reduce CALLDATA non zero bytes from 68 to 16 as https://eips.ethereum.org/EIPS/eip-2028
- Related to https://github.com/gnosis/safe-relay-service/issues/193


gnosis/safe/safe_creation_tx.py
```diff
@@ -134,14 +132,14 @@ def _calculate_gas(owners: List[str], safe_setup_data: bytes, payment_token: str
 
         base_gas = 60580  # Transaction standard gas
 
-        # TODO If we already have the token, we don't have to pay for storage, so it will be just 5K instead of 20K.
+        # If we already have the token, we don't have to pay for storage, so it will be just 5K instead of 20K.
         # The other 1K is for overhead of making the call
         if payment_token != NULL_ADDRESS:
-            payment_token_gas = 21000
+            payment_token_gas = 55000
```
rmeissner: 55k? Is that on purpose?

Uxio0: Yes, it was working because of the overestimation. First ERC20 transfer is very expensive, and 55K is what we used in CREATE2 creation, so I've just used the same

rmeissner: For me this feels extremely high and is completely counter intuitive when reading the comment above. We can assume that we normally own some tokens of the possible payment tokens.

This function contains a lot of magic values that are left unexplained (e.g. `base_gas = 60580` ... not sure what standard gas this is, looking at the comment I would have assumed `21000`, `+ 270000` ... not sure where this is coming from). This makes reviewing this pretty hard :/

rmeissner: Ahh this is the deprecated creation flow anyways, right? In this case it is not super important

gnosis/eth/constants.py
```diff
@@ -12,3 +12,7 @@
 SIGNATURE_S_MAX_VALUE: int = secpk1n // 2
 SIGNATURE_V_MIN_VALUE: int = 27
 SIGNATURE_V_MAX_VALUE: int = 28
+
+
+CALL_DATA_ZERO_BYTE = 4
```
rmeissner: probably would prefix these 2 constants with `GAS_` or something similar

Uxio0: 👍

PR #16, opened 2020-02-11
Fix 63/64 EIP-150 problem

Uxio0:


gnosis/safe/safe.py
```diff
@@ -431,32 +431,40 @@ def estimate_tx_gas_with_web3(self, to: str, value: int, data: bytes) -> int:
         """
         Estimate tx gas using web3
         """
-        return self.ethereum_client.estimate_gas(self.address, to, value, data)
+        gas_estimated = self.ethereum_client.estimate_gas(self.address, to, value, data)
+        for _ in range(20):  # Test `call()` and increase gas 20 times, 63/64th problem
+            try:
```
rmeissner: This will not work for delegate calls right?

Uxio0: No, delegate calls are disabled for his kind of estimation: https://github.com/gnosis/gnosis-py/blob/eip-150-fix/gnosis/safe/safe.py#L460

rmeissner: it this does not throw then we just perform the same call 20 times?

Uxio0: @rmeissner You are right, I fixed it

gnosis/safe/safe.py
```diff
@@ -431,32 +431,41 @@ def estimate_tx_gas_with_web3(self, to: str, value: int, data: bytes) -> int:
...
+                self.w3.eth.call({'gas': gas_estimated, 'from': self.address,
+                                  'to': to, 'value': value,
+                                  'data': data})
+                return gas_estimated
```
Uxio0: @rmeissner It should be alright now

PR #19, opened 2020-06-24
feat: Add support for the Energy Web Chain and the Volta testnet

JosephBagaric:
Needed in order to add `Volta` and `Energy Web Chain` support to https://github.com/gnosis/safe-transaction-service for https://github.com/gnosis/safe-react/pull/1046


Uxio0:
Thanks, I will merge it. But there's no need of this for the transaction service. This line https://github.com/gnosis/safe-transaction-service/blob/master/safe_transaction_service/history/management/commands/setup_service.py#L58 can be changed to `ethereum_network_id = ethereum_client.w3.net.version` and use the `network id` instead of the `enum`

PR #20, opened 2020-09-04
Make CPK Factory Proxy valid

germartinez:
Make CPK Factory Proxy valid

- Added `CPKFactoryProxy.json`
- Updated `check_proxy_code` to take `CPKFactoryProxy.json` into account

### DIFF
--- a/gnosis/eth/contracts/CPKFactoryProxy.json
+++ b/gnosis/eth/contracts/CPKFactoryProxy.json
@@ -1,0 +1,945 @@
+{
+  "contractName": "Proxy",
+  "abi": [
+    {
+      "inputs": [
+        {
+          "internalType": "address",
+          "name": "_masterCopy",
+          "type": "address"
+        }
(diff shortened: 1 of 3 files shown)


Uxio0:
Before including this on the relay I would like to know if we have an official contract deployment for the CPK for all networks (with the same address if possible) and with the same bytecode @cag @rmeissner @germartinez

rmeissner:
So they were deployed by @cag but we do not have the factories on the same addresses (and probably also not same bytecode) in the different networks and we only need to support rinkeby and mainnet for this (and local development). Therefore it probably doesn't make sense to include the info as a "hardcoded" json file. If possible it would be nice to configure that as a service configuration ... but that might be a little bit more work then. Any estimate how hard this would be?

Uxio0:
It's is just to support 2 copies I would say to just hardcode both, it would be way easier than change the logic just for the CPK

PR #25, opened 2021-02-10
fix link

ofek:
I just moved everything to the new docs site

cc @Uxio0


Uxio0:
Thanks for your PR @ofek

PR #27, opened 2021-03-11
Refactor Safe gas estimation

Uxio0:
- Use web3 as fallback

### DIFF
--- a/gnosis/eth/ethereum_client.py
+++ b/gnosis/eth/ethereum_client.py
@@ -1081,8 +1081,6 @@ class EthereumClient:
         :return:
         """
         tx: TxParams = {'to': to}
-        if to:
-            tx['to'] = to
         if from_:
             tx['from'] = from_
         if value:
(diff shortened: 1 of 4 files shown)


gnosis/eth/ethereum_client.py
```diff
@@ -1066,61 +1066,37 @@ def batch_call(self, contract_functions: Iterable[ContractFunction], from_addres
...
-                    "to": to,
-                    "data": data,
-                    "value": Wei(value),
-                })
-            else:
-                raise ValueError(response_json['error'])
-        else:
-            return int(response_json['result'], 16)
+        tx: TxParams = {'to': to}
+        if to:
```
rmeissner, line 1084: Is this one required as you set it in the first step anyways?

rmeissner (approved):
Looks like it makes sense to me

gnosis/eth/ethereum_client.py
Uxio0: You are right, thanks

PR #32, opened 2021-05-05
Fixed fetching the default db connection in newer versions of Django.

HayesData:
Fixed import error when importing DefaultConnectionProxy in newer versions of Django as this has been replaced and is noted in the Django codebase as can be seen in the link below.

https://github.com/django/django/blob/06fd4df41afb5aa1d681b853c3c08d8c688ca3a5/django/db/__init__.py#L21


Uxio0 (approved):
Thank you very much for the fix

PR #45, opened 2021-07-23
Fix web3 version range definition

fmrsabino:
- The `web3` version range is incorrectly set so when pip downloads the dependencies it actually downloads the latest available version (which has breaking API changes)
- The version should be set according to: https://www.python.org/dev/peps/pep-0508/#grammar

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ requirements = [
     "py-eth-sig-utils>=0.3.0",
     "typing-extensions==3.10.0.0; python_version < '3.8'",
     "requests>=2",
-    "web3>=5.18<=5.19",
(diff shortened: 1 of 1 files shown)


mikheevm (approved):

setup.py
```diff
@@ -15,7 +15,7 @@
     "py-eth-sig-utils>=0.3.0",
     "typing-extensions==3.10.0.0; python_version < '3.8'",
     "requests>=2",
-    "web3>=5.18<=5.19",
+    "web3 >=5.18, <=5.19",
```
mikheevm, line 18: nitpick: other package don't have a space between name/version

mikheevm (approved):
🎉

PR #57, opened 2021-08-23
Bump web3 version to 5.23.0 in setup.py

fmrsabino:
- Set `web3` to version `5.23.0` as a requirement.
- Current dependents on this library that update to version `3.2.0` download an incompatible version of `web3`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ requirements = [
     "py-eth-sig-utils>=0.3.0",
     "typing-extensions==3.10.0.0; python_version < '3.8'",
     "requests>=2",
-    "web3>=5.18,<=5.19",
+    "web3==5.23.0",
 ]
 
 extras_require = {


fmrsabino:
Please do review 4d468e2 separately. If this solution is agreed on I can open a PR with it and then we can merge this one

fmrsabino:
> Please do review [4d468e2](https://github.com/gnosis/gnosis-py/commit/4d468e263c9a4208c9b910753f9e66be01ee82c1) separately. If this solution is agreed on I can open a PR with it and then we can merge this one

Decided to open a PR with this change: https://github.com/gnosis/gnosis-py/pull/58

setup.py
```diff
@@ -15,7 +15,7 @@
     "py-eth-sig-utils>=0.3.0",
     "typing-extensions==3.10.0.0; python_version < '3.8'",
     "requests>=2",
-    "web3>=5.18,<=5.19",
+    "web3==5.23.0",
```
jpalvarezl, line 18: The code previously was specifying that `web3` should be either `5.18.x` or `5.19.x`. Is it totally safe to update this?

fmrsabino: Yes this update was part of this change https://github.com/gnosis/gnosis-py/pull/43 which was released here https://github.com/gnosis/gnosis-py/releases/tag/v3.2.0

So the project itself requires `5.23.0` as a dependency but this dependency has an incompatible version being exposed to projects who use `gnosis-py`. So:
- For gnosis-py `web3 5.23.0` is already set and being used
- For projects that depend on the gnosis-py they fetch `web>=5.18,<=5.19` which is incompatible with the latest release (example: https://github.com/gnosis/safe-config-service/pull/198 – check actions tab for this change)

jpalvarezl (approved):

PR #58, opened 2021-08-23
Update cache step. Remove Github Action venv.

fmrsabino:
- Due to some caching issues (skipping dependency installation when required) the cache action of the `pip` dependencies was updated – this should result in cache hits when the dependencies are already available under `~/.cache/pip`
- Remove `venv` – this should make it easier to maintain the cached dependencies by the Github Actions (as now they are installed under `~/.cache/pip`

Example of action that was failing due to cache issues: https://github.com/gnosis/gnosis-py/actions/runs/1159158074

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -21,8 +21,7 @@ jobs:
           - 5432:5432
     steps:
     - name: Setup and run ganache
-      run: |
-        docker run --detach --publish 8545:8545 --network-alias ganache -e DOCKER=true trufflesuite/ganache-cli:latest --defaultBalanceEther 10000 --gasLimit 10000000 -a 30 --noVMErrorsOnRPCResponse --chainId 1337 --networkId 1337 -d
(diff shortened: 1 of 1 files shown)


fmrsabino:
@jpalvarezl changes to single line run commands here: d3c41a1

.github/workflows/python.yml
```diff
@@ -21,46 +21,34 @@ jobs:
...
         python-version: ${{ matrix.python-version }}
-    - name: Create virtualenv
-      run: |
-        python -m venv venv
     - uses: actions/cache@v2
-      id: cache-pip
       with:
-        path: ./venv
-        key: ${{ matrix.python-version }}-${{ hashFiles('requirements-test.txt') }}-${{ hashFiles('requirements.txt') }}
+        path: ~/.cache/pip
```
jpalvarezl, line 32: The cached path is different. Is that ok?

jpalvarezl (approved):
Just one comment on a change that is probably meant on purpose. I approve so as to not block you. Looks good 👍

.github/workflows/python.yml
fmrsabino: Yes. The virtual environment is gone so `/venv` no longer exists.

`~/.cache/pip` is the default location on ubuntu (where this action runs)

PR #67, opened 2021-09-02
Add Polis Olympus (mainnet)

eabz:
Hello, we recently launched our mainnet, Gnosis is critical for our DAO system and consensus.

Polis is an AuRa based blockchain using POSDAO.

Adding to this library will help us simplify our structure.


Uxio0 (approved):
Sure, thanks for your PR and good luck with the project!

PR #68, opened 2021-09-03
Add support for MakerDAO's multicall contract

Uxio0:


gnosis/eth/multicall.py
```diff
@@ -0,0 +1,182 @@
...
+        tx_receipt = ethereum_client.get_transaction_receipt(tx_hash, timeout=120)
+        assert tx_receipt and tx_receipt['status']
+        contract_address = tx_receipt['contractAddress']
+        logger.info("Deployed Multicall V2 Contract %s by %s", contract_address,
+                    deployer_account.address)
+        # Add address to addresses dictionary
+        cls.ADDRESSES[ethereum_client.get_network()] = contract_address
+        return EthereumTxSent(tx_hash, tx, contract_address)
+
+    def _build_payload(self, contract_functions: Sequence[ContractFunction]
```
fmrsabino, line 94: 💡  `self` seems to not be used here. maybe we can convert this method to a `@staticmethod`

gnosis/eth/tests/ethereum_test_case.py
```diff
@@ -8,22 +8,38 @@
...
     def setUpClass(cls):
         super().setUpClass()
-        cls.ethereum_client = EthereumClientProvider()
-        cls.w3 = cls.ethereum_client.w3
+
         cls.ethereum_test_account = Account.from_key(settings.ETHEREUM_TEST_PRIVATE_KEY)
+        # Caching ethereum_client to prevent initializing again
+        cls.ethereum_client = _cached_data['ethereum_client']
+
+        if not cls.ethereum_client:
```
fmrsabino, line 36: ❓Would `@functools.cached_property` work here? It'd would work on the lifecycle of the `EthereumTestCaseMixin` instance but a new instance would initialise it again right? 🤔

gnosis/eth/tests/ethereum_test_case.py
```diff
@@ -8,22 +8,38 @@
...
-    w3: Web3
-    ethereum_test_account: LocalAccount
+    ethereum_client: EthereumClient = None
+    w3: Web3 = None
+    ethereum_test_account: LocalAccount = None
+    multicall: Multicall = None
 
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
```
fmrsabino, line 30: ⚠️ There's no `super()` here to be called 🤔

fmrsabino (changes requested):
Awesome! 🚀 

Mostly questions that I have.

I noticed that `gnosis/py.typed` got removed but it is still referenced by `setup.py`

gnosis/eth/ethereum_client.py
```diff
@@ -33,6 +33,19 @@
 from .typing import BalanceDict, EthereumData, EthereumHash
 from .utils import decode_string_or_bytes32
 
+try:
+    from functools import cached_property
+except ImportError:
+    from cached_property import cached_property
```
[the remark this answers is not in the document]
Uxio0: That's right

gnosis/eth/multicall.py
Uxio0: Agree

gnosis/eth/tests/ethereum_test_case.py
Uxio0: That's right. And we are using `EthereumTestCaseMixin` in multiple places

gnosis/eth/tests/ethereum_test_case.py
Uxio0: Well, that's because it's a `mixin` and not a "proper" class. It's intended to be inherited with another class (a `TestCase` in this case)

fmrsabino: Got it 🙌

fmrsabino (approved):
Nice 🚀 🚀 🚀

PR #91, opened 2021-10-01
Add price oracle for curve gauge deposits

mikheevm:
Required to fix https://github.com/gnosis/safe-transaction-service/issues/302#issuecomment-880714463

https://resources.curve.fi/base-features/understanding-gauges

The Curve adapter doesn't recognize the token if it's a gauge deposit. So I added a dictionary of gauge deposit addresses to the Vault LP tokens. I retrieved it from Curve API https://api.curve.fi/api/getGauges. So now it checks if an address is a gauge deposit and replaces the address with an LP token address in such a case.


gnosis/eth/tests/test_oracles.py
```diff
@@ -360,7 +374,7 @@ def test_get_underlying_tokens(self):
         self.assertEqual(underlying_token.address, mln_vault_underlying_token)
 
         underlying_tokens = enzyme_oracle.get_underlying_tokens(usf_fund_token_address)
-        self.assertEqual(len(underlying_tokens), 5)
+        self.assertEqual(len(underlying_tokens), 4)
```
mikheevm, line 377: they rebalanced the fund so this failed

Uxio0 (approved):
Really nice job!

PR #97, opened 2021-10-04
Run pre-commit hook on CI

Uxio0:


requirements-test.txt
```diff
@@ -1,5 +1,5 @@
 -r requirements.txt
-coverage==6.0
+coverage==5.5
```
jpalvarezl, line 2: Just out of curiosity, why the downgrade?

jpalvarezl (approved):

mikheevm (approved):

requirements-test.txt
Uxio0: Because new version does not seem to be compatible with `coveralls` (yet)

PR #120, opened 2021-11-03
feat: expose safe tx eip712 structured data

banteg:
Expose the EIP712 structured data of a SafeTx as `safe_tx_data` (extracted from `safe_tx_hash`).

This structure could be useful elsewhere, ranging from third-party signing to verification purposes.

My use case is passing it to Frame for signing.

### DIFF
--- a/gnosis/safe/safe_tx.py
+++ b/gnosis/safe/safe_tx.py
@@ -138,7 +138,7 @@ class SafeTx:
             return self.contract.functions.VERSION().call()
 
     @property
-    def safe_tx_data(self) -> Dict:
+    def eip712_structured_data(self) -> Dict:
         data = self.data.hex() if self.data else ""
 
         # Safes >= 1.0.0 Renamed `baseGas` to `dataGas`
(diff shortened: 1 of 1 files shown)


gnosis/safe/safe_tx.py
```diff
@@ -138,7 +138,7 @@ def safe_version(self) -> str:
             return self.contract.functions.VERSION().call()
 
     @property
-    def safe_tx_hash(self) -> HexBytes:
+    def safe_tx_data(self) -> Dict:
```
Uxio0, line 141: I think this name is misleading. What do you think renaming the function to `def eip712_structured_data(self)`?

Uxio0 (changes requested):
Good PR! I'm also thinking on migrating to the newer library `eip712_structs` that I'm using for the Gnosis Protocol: https://github.com/gnosis/gnosis-py/blob/master/gnosis/protocol/order.py

banteg:
There is also a good library called [eip712](https://github.com/ApeWorX/eip712/blob/68f5ebbf8603dcd137251eed93ace2caaed09e2d/tests/test_messages.py#L56-L66), but I found a [name clash bug](https://github.com/ApeWorX/eip712/issues/9) in it when trying to implement the struct used here.

Uxio0 (approved):

PR #121, opened 2021-11-04
Port py-eth-sig-utils to eip712-structs

banteg:
Followed @Uxio0 suggestion from here https://github.com/gnosis/gnosis-py/pull/120#pullrequestreview-796306506

Rewrote EIP712 code of SafeTx to match the library used in Protocol. Removed `py-eth-sig-utils` dependency.

### DIFF
--- a/gnosis/safe/safe_tx.py
+++ b/gnosis/safe/safe_tx.py
@@ -42,6 +42,36 @@ except ImportError:
     from cached_property import cached_property
 
 
+class EIP712SafeTx(EIP712Struct):
+    to = Address()
+    value = Uint(256)
+    data = Bytes()
+    operation = Uint(8)
+    safeTxGas = Uint(256)
+    baseGas = Uint(256)  # `dataGas` was renamed to `baseGas` in 1.0.0
+    gasPrice = Uint(256)
+    gasToken = Address()
+    refundReceiver = Address()
+    nonce = Uint(256)
+
+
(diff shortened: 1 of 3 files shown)


gnosis/safe/safe_tx.py
```diff
@@ -138,62 +139,67 @@ def safe_version(self) -> str:
...
-        # Safes >= 1.3.0 Added `chainId` to the domain
-        if safe_version >= Version("1.3.0"):
-            # EIP712Domain(uint256 chainId,address verifyingContract)
-            structured_data["types"]["EIP712Domain"].insert(
-                0, {"name": "chainId", "type": "uint256"}
-            )
-            structured_data["domain"]["chainId"] = self.chain_id
+        if safe_version >= Version("1.0.0"):
+
+            class SafeTx(EIP712Struct):
```
Uxio0, line 148: I wouldn't define this class in an `if`, I think it's better to do it at the beginning of the `safe_tx.py` file. Then define 2 classes, `LegacySafeTx` and `SafeTx`

Uxio0 (changes requested):
Very nice PR, what do you think about the change I proposed?

gnosis/safe/safe_tx.py
banteg: Makes sense, pushed this change.

Uxio0 (approved):
Nice PR, thanks!

PR #129, opened 2021-11-12
Add `chainId` to EIP 1559 txs

Uxio0:
- `chainId` is mandatory for eip1559 txs
- Rename `eip1159` -> `eip1559`

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -51,7 +51,6 @@ jobs:
     - name: Test setup.py
       run: pip install -e .
     - name: Send results to coveralls
-      if: ${{ env.COVERALLS_REPO_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/python.yml
```diff
@@ -51,7 +51,6 @@ jobs:
     - name: Test setup.py
       run: pip install -e .
     - name: Send results to coveralls
-      if: ${{ env.COVERALLS_REPO_TOKEN }}
```
fmrsabino, line 54: Is this change intentional?

fmrsabino (approved):

.github/workflows/python.yml
Uxio0: Yes, coveralls where not running

PR #135, opened 2021-11-17
Set timeout for all EthereumClient http requests

Uxio0:


gnosis/eth/ethereum_client.py
```diff
@@ -1286,8 +1298,10 @@ def _prepare_http_session(self) -> requests.Session:
         session = requests.Session()
         adapter = requests.adapters.HTTPAdapter(
             pool_connections=1,  # Doing all the connections to the same url
-            pool_maxsize=100,
-        )  # Do many requests to the same host
+            pool_maxsize=100,  # Do many requests to the same host
```
fmrsabino, line 1301: A bit confused by the comment. Here it means that this client supports at most `100` concurrent connections right?

fmrsabino (approved):
🚀

gnosis/eth/ethereum_client.py
Uxio0: Yes; I will change it

PR #149, opened 2021-12-01
Add basic ERC1155 support.

nepeat:
Manager not implemented but this adds the ABI to simplify basic ERC1155 operations like ERC721.


Uxio0:
Hi and thanks for the PR. What ERC1155 did you use? As you understand I don't want to merge bytecode without checking. If it's not easy for you I can just compile a example ERC1155 from Openzeppelin and add it myself, would that work for you?

Uxio0:
I see, OpenZeppelin. I got the same bytecode compiling now. Merging!

Uxio0 (approved):

PR #195, opened 2022-02-16
Fix decoding ERC20/721 Transfer events

Uxio0:
- If a not standard event was used, like `Transfer(address indexed, address, uint256 indexed)`, decoding was broken

### DIFF
--- a/gnosis/eth/tests/test_ethereum_client.py
+++ b/gnosis/eth/tests/test_ethereum_client.py
@@ -6,7 +6,6 @@ from django.test import TestCase
 
 from eth_account import Account
 from hexbytes import HexBytes
-from web3.datastructures import AttributeDict
 from web3.eth import Eth
 from web3.types import TxParams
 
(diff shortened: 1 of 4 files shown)


gnosis/eth/tests/test_ethereum_client.py
```diff
@@ -34,281 +34,10 @@
...
-                    "blockHash": HexBytes(
-                        "0xdf3c33d034f1b342820afdfb2612d0794d4a1c15518184d71a047ef1eb151d10"
-                    ),
-                    "logIndex": 5,
-                    "removed": False,
-                }
-            ),
-        ]
-
         # TODO Test with ganache
```
fmrsabino, line 37: Is this still valid?

fmrsabino (commented):
Looks good! I wonder if it makes more sense if we see a `DecodingError` to propagate that from `decode_logs`

fmrsabino (approved):

Uxio0:
> Looks good! I wonder if it makes more sense if we see a `DecodingError` to propagate that from `decode_logs`

That will change the current logic. Currently if something cannot be decoded `None` is returned and an warning is logged. Btw I used `logging.error` by mistake, I switched it to `logging.warning`

PR #215, opened 2022-03-24
Ignore invalid padding for Safe tx signatures

Uxio0:
Padding bytes are ignored by Solidity, but not by Python, so I trim them and replace by null bytes

### DIFF
--- a/gnosis/safe/safe_signature.py
+++ b/gnosis/safe/safe_signature.py
@@ -6,6 +6,7 @@ from typing import List, Union
 from eth_abi import decode_single, encode_single
 from eth_abi.exceptions import DecodingError
 from eth_account.messages import defunct_hash_message
+from eth_typing import ChecksumAddress
 from eth_utils import to_checksum_address
 from hexbytes import HexBytes
 from web3.exceptions import BadFunctionCallOutput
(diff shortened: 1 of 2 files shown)


gnosis/safe/safe_signature.py
```diff
@@ -50,6 +51,19 @@ def from_v(v: int):
...
+def uint_to_address(value: int) -> ChecksumAddress:
+    """
+    Convert a Solidity `uint` value to a checksummed `address`, removing
+    invalid padding bytes if present
+
+    :return: Checksummed address
+    """
+    encoded = encode_single("uint", value)
+    # Remove padding bytes, as Solidity will ignore it but `eth_abi` will not
+    encoded_without_padding_bytes = b"\x00" * 12 + encoded[-20:]
```
rmeissner, line 63: I would have expected the python lib to handle this.

rmeissner (approved):

fmrsabino (approved):

gnosis/safe/safe_signature.py
Uxio0: Yeah, I would too 😢 : https://github.com/ethereum/eth-abi/issues/162

PR #219, opened 2022-03-28
fix chain id

Uxio0:
- Fix chainId for not existing networks on our enum
- Set version 3.9.1

### DIFF
--- a/gnosis/eth/ethereum_client.py
+++ b/gnosis/eth/ethereum_client.py
@@ -1339,13 +1339,20 @@ class EthereumClient:
         return self.w3.eth.block_number
 
     @cache
+    def get_chain_id(self) -> int:
+        """
+        :return: ChainId returned by the RPC `eth_chainId` method. It should never change, so it's cached.
+        """
+        return int(self.w3.eth.chain_id)
+
     def get_network(self) -> EthereumNetwork:
         """
-        Get network name based on the network version id. It should never change, so it's cached
+        Get network name based on the chainId
 
-        :return: The EthereumNetwork enum type
+        :return: EthereumNetwork based on the chainId. If network is not
+            on our list, `EthereumNetwork.UNKOWN` is returned
         """
-        return EthereumNetwork(int(self.w3.eth.chain_id))
+        return EthereumNetwork(self.get_chain_id())
 
     @cache
     def is_eip1559_supported(self) -> EthereumNetwork:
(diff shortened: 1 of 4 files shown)


fmrsabino (commented):
Wouldn't we still have partial functionality in the end?

- `get_chain_id` would work for any chain
- `get_network` would only work with our enum

Uxio0:
> Wouldn't we still have partial functionality in the end?
> 
> * `get_chain_id` would work for any chain
> * `get_network` would only work with our enum

Yes, I don't like it either... Do you have any idea?

fmrsabino:
> > Wouldn't we still have partial functionality in the end?
> > 
> > * `get_chain_id` would work for any chain
> > * `get_network` would only work with our enum
> 
> Yes, I don't like it either... Do you have any idea?

I think if the core functionality relies on the supported networks of the enum then I would still use the enum and the chain id still needs to be in the enum.

I think it's fine if we say that the networks that we support were part of a review process so it shouldn't work with any `chain_id` but with the ones that we reviewed (since we have dependencies like block explorers, and other types of metadata).

The alternative would be to refactor anything to not take the enum into consideration and work with the chain id while taking the metadata that we have right now in the repo as optional.

Uxio0:
> I think if the core functionality relies on the supported networks of the enum then I would still use the enum and the chain id still needs to be in the enum.

Then you are breaking people trying to use this library (or even the safe-cli) on private networks

> I think it's fine if we say that the networks that we support were part of a review process so it shouldn't work with any `chain_id` but with the ones that we reviewed (since we have dependencies like block explorers, and other types of metadata).

I disagree, we should try to support every network possible without our intervention.

> The alternative would be to refactor anything to not take the enum into consideration and work with the chain id while taking the metadata that we have right now in the repo as optional.

I think it's alright if we use the `get_chain_id` for basic operation and `get_network` when we need to work based on networks that exist (like checking the balances or configuring stuff depending on the network)

fmrsabino:
> Then you are breaking people trying to use this library (or even the safe-cli) on private networks

We are not as this is the current behaviour or was this issue introduced recently? 🤔 

> I disagree, we should try to support every network possible without our intervention.

That's what I meant: "I think it's fine **if** we say" – this is to say that depending on the scope of what we want to support then I can understand this change.

So if it's clear that we want to support _any_ chain then I think we should say what might not work if the chain is not part of the enum.

> I think it's alright if we use the get_chain_id for basic operation and get_network when we need to work based on networks that exist (like checking the balances or configuring stuff depending on the network)

Also fine with me. I think it just needs to be clear our level of support because before we would not support chains not available under the enum.

fmrsabino (approved):
As I mentioned in https://github.com/gnosis/gnosis-py/pull/219#issuecomment-1080466235

For me this change is fine although it seems to be more like a major release in terms of support than an hotfix.

But this is just my personal take on it.

Uxio0:
> As I mentioned in [#219 (comment)](https://github.com/gnosis/gnosis-py/pull/219#issuecomment-1080466235)
> 
> For me this change is fine although it seems to be more like a major release in terms of support than an hotfix.
> 
> But this is just my personal take on it.

For me it's only a hotfix. There's no change on any function or signature

PR #222, opened 2022-03-30
Update pre-commit

Uxio0:


.pre-commit-config.yaml
```diff
@@ -6,15 +6,15 @@ repos:
     hooks:
       - id: isort
   - repo: https://github.com/psf/black
-    rev: 22.1.0
+    rev: 22.3.0
     hooks:
       - id: black
```
fmrsabino, line 11: We have this pre-commit hook but it seems that we don't declare `black` as a dependency.

Should we provide this in `requirements-dev.txt`

fmrsabino (approved):

.pre-commit-config.yaml
Uxio0: Pre-commit installs its packages independently

fmrsabino: Yes under a pre-commit environment. It's still not available for in the normal `venv` right?

Uxio0: No

fmrsabino: > No

Ok so we don't add `black` to the dependencies because it's under the pre-commit.
I would say that we can remove `flake8` and `isort` from `requirements-dev.txt` so that we don't have any conflict with the versions.

Uxio0: I think we can remove `isort`, but `flake8` is used by the editor (like vim or vscode). Thinking about it, I think we can add black to get auto formatting in the IDE, what do you think?

fmrsabino: > I think we can remove `isort`, but `flake8` is used by the editor (like vim or vscode). Thinking about it, I think we can add black to get auto formatting in the IDE and not remove anything, what do you think?

I think that's fine 👍 

Was just wondering about the value of having those in `requirements-dev.txt` and also in the pre-commit. In the safe-config-service we have on both (the disadvantage is that the versions need to be kept in sync) 🤔

PR #235, opened 2022-04-21
Cache some web3 queries that shouldn't change

Uxio0:
- Some web3 queries never change and can be cached: `chainId`,
  `getBlockByHash`...
- Use [web3 simple cache middleware](https://web3py.readthedocs.io/en/stable/middleware.html#web3.middleware.construct_simple_cache_middleware)

### DIFF
--- a/gnosis/eth/ethereum_client.py
+++ b/gnosis/eth/ethereum_client.py
@@ -38,7 +38,7 @@ from web3.exceptions import (
     TimeExhausted,
     TransactionNotFound,
 )
-from web3.middleware import geth_poa_middleware
+from web3.middleware import geth_poa_middleware, simple_cache_middleware
 from web3.types import (
     BlockData,
     BlockIdentifier,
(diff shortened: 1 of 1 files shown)


gnosis/eth/ethereum_client.py
```diff
@@ -38,7 +38,7 @@
     TimeExhausted,
     TransactionNotFound,
 )
-from web3.middleware import geth_poa_middleware
+from web3.middleware import geth_poa_middleware, simple_cache_middleware
```
fmrsabino, line 41: Interesting that they provide this.

It's not clear to me the default behaviour of this middleware though. They do mention in the docs that caching will apply to any call present in `rpc_whitelist`. Do we need to build this `rpc_whitelist`?

https://web3py.readthedocs.io/en/stable/middleware.html#web3.middleware.construct_simple_cache_middleware

Uxio0: No, there's a default list of methods to be cached that shouldn't change: https://github.com/ethereum/web3.py/blob/master/web3/middleware/cache.py#L35

fmrsabino: Ah yes. I see the default argument now! Thanks.

fmrsabino (approved):
🚀

PR #242, opened 2022-05-10
Fix slow timeout typo

Uxio0:
- Fix slow timeout timeout for EthereumClient
- Set version 3.9.4


luarx (approved):

fmrsabino (approved):
Just to be sure: it was not a typo but just missing setting the `timeout` parameter right?

PR #249, opened 2022-05-16
Rename package from gnosis-py to safe-eth-py

Uxio0:


rmeissner (approved):

Uxio0:
`gnosis-py` was also updated to point to this package: https://pypi.org/project/gnosis-py/

PR #269, opened 2022-06-15
Optimize `to_checksum_address` function

Uxio0:
While developing Safe Tx Service we found out that a lot of time was spent on converting addresses from `bytes` to `ChecksumAddress`. Biggest part of this time was overhead added on `eth_utils` library to `keccak256` function.

So we decided to use `pysha3` library (way faster than `pycryptodome`) directly. That's why we implemented new methods:

- fast_keccak
- fast_keccak_hex
- fast_to_checksum_address
- fast_bytes_to_checksum_address
- fast_is_checksum_address

This is also very relevant for the `EthereumAddressV2Field`, as it stores addresses as `bytes` on database and returns them as `ChecksumAddress`

Related to:
  - https://github.com/ethereum/eth-utils/issues/95
  - https://github.com/ethereum/eth-hash/issues/35

### DIFF
--- a/gnosis/eth/utils.py
+++ b/gnosis/eth/utils.py
@@ -1,12 +1,103 @@
+import warnings
 from secrets import token_bytes
 from typing import Tuple, Union
 
 import eth_abi
 from eth._utils.address import generate_contract_address
 from eth_keys import keys
-from eth_utils import to_canonical_address, to_checksum_address
+from eth_typing import AnyAddress, ChecksumAddress, HexStr
+from eth_utils import to_normalized_address
 from hexbytes import HexBytes
-from web3 import Web3
+from sha3 import keccak_256
+
+
+def fast_keccak(value: bytes) -> bytes:
+    """
+    Calculates ethereum keccak256 using fast library `pysha3`
+    :param value:
+    :return: Keccak256 used by ethereum as `bytes`
+    """
+    return keccak_256(value).digest()
+
+
+def fast_keccak_hex(value: bytes) -> HexStr:
+    """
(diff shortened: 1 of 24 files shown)


Uxio0:
Related to https://github.com/safe-global/safe-pm/issues/89

gnosis/eth/utils.py
```diff
@@ -4,9 +4,97 @@
...
+    :return:
+    """
+    return ChecksumAddress(
+        "0x"
+        + (
+            "".join(
+                (
+                    norm_address[i].upper()
+                    if int(address_hash[i], 16) > 7
+                    else norm_address[i]
```
mikhailxyz, line 48: I'd add a link to https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md as a comment

gnosis/safe/safe.py
```diff
@@ -888,9 +892,9 @@ def retrieve_fallback_handler(
             self.address,
             self.FALLBACK_HANDLER_STORAGE_SLOT,
             block_identifier=block_identifier,
-        )[-20:]
+        )[-20:].rjust(20, b"\0")
```
moisses89, line 895: Why needs left padding?

Uxio0: Because `get_storage` ignores zeros, don't really know why

gnosis/eth/utils.py
Uxio0: Nice, done @mikhailxyz

mikhailxyz (approved):

moisses89 (approved):

PR #292, opened 2022-07-07
Add classes to interact with Safe APIs

### LINKED ISSUE #284: Introduce method `SafeTx.api_post_data`
Currently, in order to post Safe transactions to the web interface (with this project as a dependency), we have to manually construct the `post_data` dictionary out of all the fields already contained in `safe_tx`. Since a `SafeTx` already contains the relevant info to build the post_data, it would be nice if there were some exposed method that would return the required format.

For example, the content of this method (taken from `safe-cli`) which constructs the post data would be nice to include here in this project:

https://github.com/5afe/safe-cli/blob/7a70a18c36b163ee91d5c763208ba1b1c9571317/safe_cli/api/transaction_service_api.py#L198-L222

Uxio0:
- Support Tx Service
- Support Relay Service (deprecated)
- Closes #284

### DIFF
--- a/gnosis/safe/api/base_api.py
+++ b/gnosis/safe/api/base_api.py
@@ -4,7 +4,11 @@ from urllib.parse import urljoin
 
 import requests
 
-from gnosis.eth.ethereum_client import EthereumClient, EthereumNetwork
+from gnosis.eth.ethereum_client import (
+    EthereumClient,
+    EthereumNetwork,
+    EthereumNetworkNotSupported,
+)
 
 
 class SafeAPIException(Exception):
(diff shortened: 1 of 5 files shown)


gnosis/safe/tests/api/test_transaction_service_api.py
```diff
@@ -0,0 +1,77 @@
...
+from gnosis.eth.tests.ethereum_test_case import EthereumTestCaseMixin
+
+from ...api.transaction_service_api import TransactionServiceApi
+
+
+class TestTransactionServiceAPI(EthereumTestCaseMixin, unittest.TestCase):
+    def setUp(self) -> None:
+        self.transaction_service = TransactionServiceApi(
+            self.ethereum_client, EthereumNetwork.RINKEBY
+        )
```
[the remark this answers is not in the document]
Uxio0: That's right

gnosis/safe/api/base_api.py
```diff
@@ -0,0 +1,46 @@
+from abc import ABC
+from typing import Dict, Optional
+from urllib.parse import urljoin
+
+import requests
+
+from gnosis.eth.ethereum_client import EthereumClient, EthereumNetwork
```
bh2smith, line 7: I thought you didn't like this kind of import?

Uxio0: You are right, missed that 👍🏻

gnosis/safe/api/base_api.py
```diff
@@ -0,0 +1,46 @@
...
+        self.ethereum_client = ethereum_client
+        self.base_url = self.URL_BY_NETWORK[network]
+
+    @classmethod
+    def from_ethereum_client(
+        cls, ethereum_client: EthereumClient
+    ) -> Optional["SafeBaseAPI"]:
+        ethereum_network = ethereum_client.get_network()
+        if ethereum_network in cls.URL_BY_NETWORK:
+            return cls(ethereum_network, ethereum_client=ethereum_client)
```
fmrsabino, line 30: I was thinking about moving this to the `__init__`. Do we consider `SafeBaseAPI` to be in a valid state if the result from `from_ethereum_client` is `None`?

Uxio0: No, it's not, we should raise an error. Problem is that if we move that to `__init__` we need to make `EthereumNetwork` optional. WDYT?

bh2smith: I personally like the idea of having these network parameters optional since they are not needed to encode transactions. And there are some use cases which only want to make use of transaction encoding.

Uxio0: Added an exception if network is not found and the possibility of setting a custom base URL

fmrsabino (approved):
🚀

PR #293, opened 2022-07-07
Lock py-evm==0.5.0a3

Uxio0:
- Fix issues with dependency resolving

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,7 @@ os.chdir(os.path.normpath(os.path.join(os.path.abspath(__file__), os.pardir)))
 requirements = [
     "cached-property>=1.5; python_version < '3.8'",
     "eip712_structs",
+    "packaging",
     "py-evm==0.5.0a3",
     "pysha3>=1.0.0",
     "requests>=2",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -11,8 +11,7 @@
 requirements = [
     "cached-property>=1.5; python_version < '3.8'",
     "eip712_structs",
-    "packaging",
```
fmrsabino, line 14: So this was not required?

gnosis/eth/constants.py
```diff
@@ -1,4 +1,6 @@
-from eth.constants import SECPK1_N
+SECPK1_N = (
+    115792089237316195423570985008687907852837564279074904382605163141518161494337
+)
```
fmrsabino, line 3: `py-evm.usages--` 😛

fmrsabino (approved):

setup.py
Uxio0: It should be included in any django distrution, but we are using it, I will put it back

Uxio0: Done

PR #299, opened 2022-07-11
Detect MultiSend addresses

### LINKED ISSUE #283: Make Multisend constructor `address` field Optional
Every time I want to use MultiSend, I always have to spend several minutes trying to find the contract address. My process is usually the following:

1. [Fail] Go to https://github.com/gnosis/safe-contracts and look for a `networks.json` file.
2. [Fail] Go to etherscan and search Multisend.
3. Go to a known safe and find a MultiSend transaction and copy the `to` address.

There are a few different versions of this contract deployed and I assume it is not being redeployed very often. For convenience, I think that the address field here should be take the following form:

```
address: Optional[str] = None
```

followed by 
```py
if not address:
    address = "0x40A2aCCbd92BCA938b02010E17A5b8929b49130D"
```

Also, at the same time, for consistency, the following assertion could be removed if we used `ChecksumAddress`

```
assert fast_is_checksum_address(address), ( 
             "%s proxy factory address not valid" % address 
         ) 
```

https://github.com/safe-global/safe-eth-py/blob/791ca1ecfb3337825266d41d5f4534d7c0fe0fc0/gnosis/safe/multi_send.py#L165-L175

Uxio0:
- Based on https://github.com/safe-global/safe-deployments/blob/main/src/assets/v1.3.0/multi_send.json
- Closes #283


bh2smith (approved):
Looks like it does the trick! Would be really cool if the `build_tx_data` was reachable without a real `NODE_URL`.

I suppose if the contract ABI were exported I could just encode the function call with a dummy web3 on my own.

gnosis/safe/multi_send.py
```diff
@@ -164,16 +165,35 @@ def _decode_multisend_old_transaction(
...
+    ):
 
         self.address = address
         self.ethereum_client = ethereum_client
         self.w3 = ethereum_client.w3
 
+        if address:
+            assert fast_is_checksum_address(address), (
+                "%s proxy factory address not valid" % address
+            )
```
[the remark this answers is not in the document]
Uxio0: I'm already checking if it's a `ChecksumAddress`

gnosis/safe/multi_send.py
```diff
@@ -164,16 +165,35 @@ def _decode_multisend_old_transaction(
...
+        "0xA238CBeb142c10Ef7Ad8442C6D1f9E89e07e7761",
+        "0x998739BFdAAdde7C933B942a68053933098f9EDa",
+    )
 
-    def __init__(self, address: str, ethereum_client: EthereumClient):
-        assert fast_is_checksum_address(address), (
-            "%s proxy factory address not valid" % address
-        )
+    def __init__(
+        self, ethereum_client: EthereumClient, address: Optional[ChecksumAddress] = None
```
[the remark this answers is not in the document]
Uxio0: Done

gnosis/safe/multi_send.py
```diff
@@ -164,15 +165,41 @@ def _decode_multisend_old_transaction(
 
 class MultiSend:
     dummy_w3 = Web3()
+    MULTISEND_ADDRESSES = (
+        "0xA238CBeb142c10Ef7Ad8442C6D1f9E89e07e7761",
+        "0x998739BFdAAdde7C933B942a68053933098f9EDa",
+    )
```
moisses89, line 171: There is any reason to use Tuple instead List?

gnosis/safe/tests/test_multi_send.py
```diff
@@ -1,17 +1,39 @@
...
+        with mock.patch.object(EthereumClient, "is_contract", return_value=True):
+            multisend = MultiSend(self.ethereum_client)
+            self.assertEqual(multisend.address, multisend.MULTISEND_ADDRESSES[0])
+
+        random_address = Account.create().address
+        multisend = MultiSend(self.ethereum_client, address=random_address)
+        self.assertEqual(multisend.address, random_address)
+
+        # Check with no ethereum_client
+        multisend = MultiSend()
```
moisses89, line 34: With no ethereum_client and without address. 
Maybe could add a test with ethereum_client without address? :thinking:

Uxio0: Take a look at line `26`

Uxio0: https://github.com/safe-global/safe-eth-py/pull/299/files/a4aa66392d769200dcd9398bb9fc005370b812b2#diff-475668e4347136854bf9d221fe10c3d534b6d19fa4b4980f2b074ffea13f3964R23

PR #309, opened 2022-07-22
Return tx-hash as part of SafeTx in SafeTxApi

Uxio0:
- SafeTx has a `tx_hash` field that should be populated with the `tx_hash` retrieved from the `tx-service`


rmeissner:
Could we add a reference why this is being implemented? E.g. who requested it in what context.

rmeissner (approved):

Uxio0:
> Could we add a reference why this is being implemented? E.g. who requested it in what context.

Sure, I added the explanation. For context, it's just I forgot to add it and I'm adding it now

PR #313, opened 2022-07-26
Fix None values for HexField

Uxio0:
- Call `strip()` after checking value is an str


hectorgomezv (approved):

gnosis/eth/django/tests/test_forms.py
```diff
@@ -52,10 +52,18 @@ def test_hex_field_form(self):
...
+        form = HexForm(data={"value": 1})
+        self.assertTrue(form.is_valid())
+        self.assertEqual(form.cleaned_data["value"], HexBytes(1))
+
         form = HexForm(data={"value": ""})
         self.assertTrue(form.is_valid())
         self.assertIsNone(form.cleaned_data["value"])
 
+        form = HexForm(data={"value": None})
+        self.assertTrue(form.is_valid())
```
moisses89, line 64: Maybe should be false?

PR #325, opened 2022-08-12
Fix memory leak cache for get_contract

### LINKED ISSUE #322: Memory leak in get_contracts function
**Describe the bug**
gnosis.safe.safe.Safe.get_contract is caching the results with @cache decorator that means that can storage memory without limit. 
```
    @cache
    def get_contract(self) -> Contract:
        v_1_3_0_contract = get_safe_V1_3_0_contract(self.w3, address=self.address)
        version = v_1_3_0_contract.functions.VERSION().call()
        if version == "1.3.0":
            return v_1_3_0_contract
        else:
            return get_safe_V1_1_1_contract(self.w3, address=self.address)
```
I noticied it working in https://github.com/safe-global/safe-transaction-service/issues/1008
**Expected behavior**
Don't storage in local memory without limits.

**Possible solutions**
- remove cache decorator
- configure time to live cache limiting the memory that could reach and defining a time to clean the cache.

moisses89:
closes #322 
- define maxsize to 1024*100 (Would be enough?)
- add cachetools to requirements
- replaces `@cache ` with `@cached` defining that the key to storage the cache is the `safe_address`
- add a test to check that `currsize` have correct values
- cache `maxsize` defined in environment var

### DIFF
--- a/gnosis/safe/safe.py
+++ b/gnosis/safe/safe.py
@@ -4,6 +4,8 @@ from enum import Enum
 from logging import getLogger
 from typing import Callable, List, NamedTuple, Optional, Union
 
+from cachetools import LRUCache, cached
+from cachetools.keys import hashkey
 from eth_account import Account
 from eth_account.signers.local import LocalAccount
 from eth_typing import ChecksumAddress
@@ -40,14 +42,6 @@ from .exceptions import (
 from .safe_create2_tx import SafeCreate2Tx, SafeCreate2TxBuilder
 from .safe_creation_tx import InvalidERC20Token, SafeCreationTx
 from .safe_tx import SafeTx
-
-try:
-    from functools import cache
-except ImportError:
-    from functools import lru_cache
-
-    cache = lru_cache(maxsize=None)
-
 
 logger = getLogger(__name__)
 
@@ -822,7 +816,7 @@ class Safe:
         threshold = self.retrieve_threshold()
         return 15000 + data_bytes_length // 32 * 100 + 5000 * threshold
 
-    @cache
+    @cached(cache=LRUCache(maxsize=1024 * 100), key=lambda self: hashkey(self.address))
     def get_contract(self) -> Contract:
         v_1_3_0_contract = get_safe_V1_3_0_contract(self.w3, address=self.address)
         version = v_1_3_0_contract.functions.VERSION().call()
--- a/gnosis/safe/safe.py
+++ b/gnosis/safe/safe.py
@@ -2,6 +2,7 @@ import dataclasses
 import math
 from enum import Enum
 from logging import getLogger
+from os import environ
 from typing import Callable, List, NamedTuple, Optional, Union
 
 from cachetools import LRUCache, cached
@@ -816,7 +817,10 @@ class Safe:
         threshold = self.retrieve_threshold()
         return 15000 + data_bytes_length // 32 * 100 + 5000 * threshold
 
-    @cached(cache=LRUCache(maxsize=1024 * 100), key=lambda self: hashkey(self.address))
+    @cached(
+        cache=LRUCache(maxsize=int(environ.get("CONTRACT_CACHE_SIZE", 1024 * 100))),
+        key=lambda self: hashkey(self.address),
+    )
     def get_contract(self) -> Contract:
         v_1_3_0_contract = get_safe_V1_3_0_contract(self.w3, address=self.address)
         version = v_1_3_0_contract.functions.VERSION().call()
--- a/gnosis/safe/tests/test_safe.py
+++ b/gnosis/safe/tests/test_safe.py
@@ -815,3 +815,21 @@ class TestSafe(SafeTestCaseMixin, TestCase):
 
         balance = self.w3.eth.get_balance(to)
         self.assertEqual(value, balance)
+
+    def test_get_contracts_cache_leak(self):
+        safe_deployed = self.deploy_test_safe(
+            owners=[self.ethereum_test_account.address]
+        )
+        Safe.get_contract.cache.clear()
+        for i in range(1, 2):
+            safe = Safe(safe_deployed.address, self.ethereum_client)
+            safe.get_contract()
+        self.assertEqual(Safe.get_contract.cache.currsize, 1)
+
+        safe_deployed = self.deploy_test_safe(
+            owners=[self.ethereum_test_account.address]
+        )
+        for i in range(1, 2):
+            safe = Safe(safe_deployed.address, self.ethereum_client)
+            safe.get_contract()
+        self.assertEqual(Safe.get_contract.cache.currsize, 2)
--- a/gnosis/safe/tests/test_safe.py
+++ b/gnosis/safe/tests/test_safe.py
@@ -816,12 +816,12 @@ class TestSafe(SafeTestCaseMixin, TestCase):
         balance = self.w3.eth.get_balance(to)
         self.assertEqual(value, balance)
 
-    def test_get_contracts_cache_leak(self):
+    def test_cache_size_get_contracts(self):
         safe_deployed = self.deploy_test_safe(
             owners=[self.ethereum_test_account.address]
         )
         Safe.get_contract.cache.clear()
-        for i in range(1, 2):
+        for i in range(0, 2):
             safe = Safe(safe_deployed.address, self.ethereum_client)
             safe.get_contract()
         self.assertEqual(Safe.get_contract.cache.currsize, 1)
@@ -829,7 +829,7 @@ class TestSafe(SafeTestCaseMixin, TestCase):
         safe_deployed = self.deploy_test_safe(
             owners=[self.ethereum_test_account.address]
         )
-        for i in range(1, 2):
+        for i in range(0, 2):
             safe = Safe(safe_deployed.address, self.ethereum_client)
             safe.get_contract()
         self.assertEqual(Safe.get_contract.cache.currsize, 2)
--- a/gnosis/safe/tests/test_safe.py
+++ b/gnosis/safe/tests/test_safe.py
@@ -833,3 +833,5 @@ class TestSafe(SafeTestCaseMixin, TestCase):
             safe = Safe(safe_deployed.address, self.ethereum_client)
             safe.get_contract()
         self.assertEqual(Safe.get_contract.cache.currsize, 2)
+
+        self.assertNotEqual(Safe.get_contract.cache.maxsize, None)
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 cached-property==1.5.2
+cachetools==5.2.0
 django==3.2.14
 django-filter==22.1
 djangorestframework==3.13.1


gnosis/safe/safe.py
```diff
@@ -822,7 +816,7 @@ def estimate_tx_operational_gas(self, data_bytes_length: int):
         threshold = self.retrieve_threshold()
         return 15000 + data_bytes_length // 32 * 100 + 5000 * threshold
 
-    @cache
+    @cached(cache=LRUCache(maxsize=1024 * 100), key=lambda self: hashkey(self.address))
```
rmeissner, line 819: Could we have this as an env var?

gnosis/safe/tests/test_safe.py
```diff
@@ -815,3 +815,21 @@ def test_send_previously_approved_tx(self):
 
         balance = self.w3.eth.get_balance(to)
         self.assertEqual(value, balance)
+
+    def test_get_contracts_cache_leak(self):
```
hectorgomezv, line 819: np: maybe the `leak` in the name indicates this wasn't working well previously, but I'm not sure it describes the unit test purpose.

gnosis/safe/tests/test_safe.py
```diff
@@ -815,3 +815,21 @@ def test_send_previously_approved_tx(self):
...
+        safe_deployed = self.deploy_test_safe(
+            owners=[self.ethereum_test_account.address]
+        )
+        Safe.get_contract.cache.clear()
+        for i in range(1, 2):
```
hectorgomezv, line 824: Maybe I'm not getting this, but this `range(1, 2)` runs just one iteration of the loop. The idea is to assure that while having N calls to `safe.get_contract()` the cache size stays at 1?

gnosis/safe/safe.py
moisses89: `maxsize`? make sense.

gnosis/safe/tests/test_safe.py
moisses89: Yes! This should starts from 0. Thanks

gnosis/safe/tests/test_safe.py
moisses89: The purpouse of the test is check that the cache storage make sense regarding the function calls. Then maybe could be better `test_cache_size_get_contracts`. What do you think?

fmrsabino: Yeah I agree that having `leak` in the test could indicate that you are testing that if it leaks or not.

The test should be more like "given a cache size of X, storing more than X would still keep the size at X size"

Then you can test the `max_size`. For testing purposes you can try having a really small `max_size` and see how it behaves.

fmrsabino:
> define maxsize to 1024*100 (Would be enough?)

What's the unit for this?

gnosis/safe/safe.py
```diff
@@ -822,7 +817,10 @@ def estimate_tx_operational_gas(self, data_bytes_length: int):
         threshold = self.retrieve_threshold()
         return 15000 + data_bytes_length // 32 * 100 + 5000 * threshold
 
-    @cache
+    @cached(
+        cache=LRUCache(maxsize=int(environ.get("CONTRACT_CACHE_SIZE", 1024 * 100))),
```
hectorgomezv, line 821: Are these parenthesis ok? I mean, maybe I'm missing something, but should the `1024 * 100` part be an argument for `environ.get()`?

moisses89: It is the default value if the environment variable is not defined.
Eg:
```
# Python program to explain os.environ object


# importing os module
import os

# Method 1
# Print the value of
# 'MY_HOME' environment variable
print("MY_HOME:", os.environ.get('MY_HOME', "Environment variable does not exist"))


# Method 2
try:
	print("MY_HOME:", os.environ['MY_HOME'])
except KeyError:
	print("Environment variable does not exist")
```

gnosis/safe/tests/test_safe.py
moisses89: @fmrsabino I was trying to add test for maxsize but I couldn't because it is defined in the decorator.   
Anyway I don't think that test maxsize is necessary because is part of the unit test on the library.

moisses89:
> > define maxsize to 1024*100 (Would be enough?)
> 
> What's the unit for this?



> > define maxsize to 1024*100 (Would be enough?)
> 
> What's the unit for this?



> > define maxsize to 1024*100 (Would be enough?)
> 
> What's the unit for this?

Reviewing the documentation, maxsize is the number of elements. Then with the default value we have for `1024*100` safe contracts, maybe is too much :) 1000 would be enough? What do you think?

hectorgomezv (approved):

gnosis/safe/tests/test_safe.py
fmrsabino: > @fmrsabino I was trying to add test for maxsize but I couldn't because it is defined in the decorator. Anyway I don't think that test maxsize is necessary because is part of the unit test on the library.

What I meant was not to test `max_size` directly but that this function uses some "max size". But probably this means that we need to read some internals of `cached` and that would kind of defeat the purpose of the test (it would be too tied to the implementation).

An example can be seen [here](https://docs.python.org/3/library/functools.html)
```python
@lru_cache(maxsize=32)
def get_pep(num):
    'Retrieve text of a Python Enhancement Proposal'
    resource = 'https://www.python.org/dev/peps/pep-%04d/' % num
    try:
        with urllib.request.urlopen(resource) as s:
            return s.read()
    except urllib.error.HTTPError:
        return 'Not Found'

>>> for n in 8, 290, 308, 320, 8, 218, 320, 279, 289, 320, 9991:
...     pep = get_pep(n)
...     print(n, len(pep))

>>> get_pep.cache_info()
CacheInfo(hits=3, misses=8, maxsize=32, currsize=8)
```

I think maybe a test could be that **it has as `maxsize`**. This way if this decorator is removed the test would fail (by using `get_contract.cache_info()`.

What do you think?

fmrsabino:
> > > define maxsize to 1024*100 (Would be enough?)
> > 
> > 
> > What's the unit for this?
> 
> Reviewing the documentation, maxsize is the number of elements. Then with the default value we have for `1024*100` safe contracts, maybe is too much :) 1000 would be enough? What do you think?

So if I'm not mistaken this Cache is tied to a specific `Safe`. A new instance of the safe would have its own `def get_contract(self) -> Contract`.

The contract behind a proxy doesn't change that often so I'd say a `max_size` of `1` should be good. The question is if a Proxy is upgraded (new master copy) how would you invalidate this? I'm assuming a new `Safe` instance with the same `address` is created.

gnosis/safe/tests/test_safe.py
moisses89: Yes it's true we could use `get_contract.cache.maxsize` but it's the same It's going to break if we remove the decorator.   
Maybe I could add a test using `tracemalloc` that takes a snapshot at the beginning of call this function and other at the end and check the memory don't pass of some value. 
This would be independent of the decorator.  :thinking:

fmrsabino: > Yes it's true we could use `get_contract.cache.maxsize` but it's the same It's going to break if we remove the decorator. Maybe I could add a test using `tracemalloc` that takes a snapshot at the beginning of call this function and other at the end and check the memory don't pass of some value. This would be independent of the decorator. 🤔

But it will break already because your test accesses the `cache` too. I think this is ok! If we remove the cache then the test should fail. The specification for this function is that it is cached and we are testing that.

My suggestion was more to test that you have a `max_size` set 👍 (ie.: it's not `None`)

moisses89:
> > > > define maxsize to 1024*100 (Would be enough?)
> > > 
> > > 
> > > What's the unit for this?
> > 
> > 
> > Reviewing the documentation, maxsize is the number of elements. Then with the default value we have for `1024*100` safe contracts, maybe is too much :) 1000 would be enough? What do you think?
> 
> So if I'm not mistaken this Cache is tied to a specific `Safe`. A new instance of the safe would have its own `def get_contract(self) -> Contract`.
> 
> The contract behind a proxy doesn't change that often so I'd say a `max_size` of `1` should be good. The question is if a Proxy is upgraded (new master copy) how would you invalidate this? I'm assuming a new `Safe` instance with the same `address` is created.
This cache is shared between instances, if we have `maxsize=10` means that we would could cached 10 `Safe(<address>).get_contract()` where each address is different.  
Regarding your question, maybe we could use TTLCache and define `ttl=60*10` (the unit are seconds). Then each 10 minutes we are going to invalidate the cache.

fmrsabino:
> This cache is shared between instances, if we have `maxsize=10` means that we could cached 10 `Safe(<address>).get_contract()` where each address is different. Regarding your question, maybe we could use [TTLCache](https://cachetools.readthedocs.io/en/latest/#cachetools.TTLCache) and define `ttl=60*10` (the unit are seconds). Then each 10 minutes we are going to invalidate the cache.

Oh ok. I just double checked and it seems that there's an instance specific cache – https://docs.python.org/3/library/functools.html#functools.cached_property

I will come back to this next Monday. Need to rethink the size of the cache.

fmrsabino:
> > This cache is shared between instances, if we have `maxsize=10` means that we could cached 10 `Safe(<address>).get_contract()` where each address is different. Regarding your question, maybe we could use [TTLCache](https://cachetools.readthedocs.io/en/latest/#cachetools.TTLCache) and define `ttl=60*10` (the unit are seconds). Then each 10 minutes we are going to invalidate the cache.
> 
> Oh ok. I just double checked and it seems that there's an instance specific cache – https://docs.python.org/3/library/functools.html#functools.cached_property
> 
> I will come back to this next Monday. Need to rethink the size of the cache.

Similar to what was mentioned here: https://github.com/safe-global/safe-eth-py/pull/325#discussion_r944407632 since this is a library that might have different use-cases I'd leave this option in a configuration file with some "sensible" default like `100`?

The Safe Transaction Service would then set this variable to one according to our needs.

fmrsabino (approved):

Uxio0:
Hi, my point of view on this:

@moisses89 , indeed you found the origin of at least some big memory issues and that's related to my understanding on how things work. I thought `Cache` was binded to every `object` but when used on a method it was binded to the `class`, and this way it only grows bigger every time a new Safe is created. Really nice work.

Regarding the implementation, I don't like too much adding `cachetools` or using a `LRUcache`, I don't think it's needed as every instance of `Safe` will only have one contract, and that memory will be released by Python as soon as we stop using the `Safe` object. I will propose a new PR handling this.

moisses89:
> and that memory will be released by Python as soon as we stop using the Safe object. 
This issue was created from this one
https://github.com/safe-global/safe-transaction-service/issues/1008#issuecomment-1210637658  
I think that when the object is created from maybe gunicorn, the memory is not released any more. what do you think @Uxio0 ?

Uxio0:
What do you think of https://github.com/safe-global/safe-eth-py/pull/339 ? I think is easier and more adequate for this case

Uxio0:
> I think that when the object is created from maybe gunicorn, the memory is not released any more. what do you think @Uxio0 ?

That's not right. As long as you stop using an object Python garbage collector `should` clean it at some point. If that's not true we would see memory leaks everywhere.

moisses89:
> What do you think of #339 ? I think is easier and more adequate for this case
Ok, I was proposing store the contract for different instances meanwhile the address was the same. In this case the idea is keep on cache while the object is live, I got it correctly?

Uxio0:
> > What do you think of #339 ? I think is easier and more adequate for this case
> 
> Ok, I was proposing store the contract for different instances meanwhile the address was the same. In this case the idea is keep on cache while the object is live, I got it correctly?

That's right

Uxio0:
Other option would be define `__eq__` and `__hash__` for the `Safe` object so a new object with the same Safe address stores the contract in the same cache, but I don't know... We can talk about it tomorrow

moisses89:
Closing this PR to continue with #339 proposal.

PR #339, opened 2022-08-24
Fix cache memory leaks

### LINKED ISSUE #322: Memory leak in get_contracts function

Uxio0:
- `@cache` is stored in the Class, not in the Object. So when multiple `Safe` objects were created, and `get_contract()` was called, cache only increased (also if the same Safe was created multiple times)
- Use now `@cached_property`, which is binded to the object, so when the object is destroyed, so is the cached value
- Closes #322
- Related to #325
- All the credit to @moisses89 who was the one realising of this behaviour

### DIFF
--- a/gnosis/safe/safe.py
+++ b/gnosis/safe/safe.py
@@ -42,11 +42,9 @@ from .safe_creation_tx import InvalidERC20Token, SafeCreationTx
 from .safe_tx import SafeTx
 
 try:
-    from functools import cache
+    from functools import cached_property
 except ImportError:
-    from functools import lru_cache
-
-    cache = lru_cache(maxsize=None)
+    from cached_property import cached_property
 
 
 logger = getLogger(__name__)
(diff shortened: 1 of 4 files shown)


moisses89:
I would like to test something like this https://github.com/safe-global/safe-eth-py/pull/325/files#diff-dbf8b4d43e58fe9c2345cff856302940182d80b34590cfd26648a7076277d8c5R819 and also check that everything is ok with the memory from tx-service.

hectorgomezv (approved):
LGTM 🙂

moisses89 (commented):
Should we continue reviewing the @cache decorator? For example this could generate other memory leak AFAIK. 
```
    @cache
        """
        :return: ChainId returned by the RPC `eth_chainId` method. It should never change, so it's cached.
        """
        return int(self.w3.eth.chain_id)
```

Uxio0:
> Should we continue reviewing the @cache decorator? For example this could generate other memory leak AFAIK.
> 
> ```
>     @cache
>     def get_chain_id(self) -> int:
>         """
>         :return: ChainId returned by the RPC `eth_chainId` method. It should never change, so it's cached.
>         """
>         return int(self.w3.eth.chain_id)
> ```

Usually EthereumClient is created as a Singleton, so this is not an issue, but we should definitely review other places

PR #342, opened 2022-08-30
Add Uniswap V3 Oracle

Uxio0:
- Add Uniswap V3 ABIs
- Add Uniswap V3 to a new file (other oracles should be refactored to 1 oracle per file)
- Remove unlimited caches for Oracles
- Refactor cache and cached_property decorator logic

### DIFF
--- a/gnosis/eth/oracles/uniswap_v3.py
+++ b/gnosis/eth/oracles/uniswap_v3.py
@@ -1,0 +1,182 @@
+import functools
+import logging
+from typing import Optional
+
+from eth_abi.exceptions import DecodingError
+from eth_typing import ChecksumAddress
+from web3.contract import Contract
+from web3.exceptions import BadFunctionCallOutput
+
+from gnosis.util import cached_property
+
+from .. import EthereumClient
+from ..constants import NULL_ADDRESS
+from ..contracts import get_erc20_contract
+from .abis.uniswap_v3 import (
+    uniswap_v3_factory_abi,
+    uniswap_v3_pool_abi,
+    uniswap_v3_router_abi,
+)
+from .oracles import CannotGetPriceFromOracle, PriceOracle
+
+logger = logging.getLogger(__name__)
+
+
+class UniswapV3Oracle(PriceOracle):
+    # https://docs.uniswap.org/protocol/reference/deployments
+    UNISWAP_V3_ROUTER = "0x68b3465833fb72A70ecDF485E0e4C7bD8665Fc45"
+
+    # Cache to optimize calculation: https://docs.uniswap.org/sdk/guides/fetching-prices#understanding-sqrtprice
+    PRICE_CONVERSION_CONSTANT = 2**192
+
+    def __init__(
+        self,
+        ethereum_client: EthereumClient,
+        uniswap_v3_router_address: Optional[ChecksumAddress] = None,
+    ):
+        """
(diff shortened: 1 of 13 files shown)


gnosis/eth/oracles/uniswap_v3.py
```diff
@@ -0,0 +1,182 @@
...
+            raise CannotGetPriceFromOracle(error_message) from e
+
+    def get_factory(self) -> Contract:
+        """
+        Factory contract creates the pools for token pairs
+
+        :return: Uniswap V3 Factory Contract
+        """
+        try:
+            factory_address = self.router.functions.factory().call()
```
moisses89, line 83: This is calling a function that contents a contract address for factory V3 of UniSwap, my question here is could be in the future this router contents the address of V4 with the next release? I want to be only sure because maybe an update could break some calls.

gnosis/util/util.py
```diff
@@ -1,5 +1,18 @@
 from typing import Any, Iterable, List
 
+try:
+    from functools import cached_property  # noqa F401
+except ImportError:
```
hectorgomezv, line 5: Out of curiosity... why do we have this fallback, is it to support old Python versions?

Uxio0: To support old python versions. It was introduced in pyhon 3.8: https://docs.python.org/3/library/functools.html#functools.cached_property

gnosis/eth/oracles/uniswap_v3.py
```diff
@@ -0,0 +1,182 @@
...
+        """
+        :param ethereum_client:
+        :param uniswap_v3_router_address:
+        :return: `True` is Uniswap V3 is available for the EthereumClient provided, `False` otherwise
+        """
+        return ethereum_client.is_contract(
+            uniswap_v3_router_address or cls.UNISWAP_V3_ROUTER
+        )
+
+    @functools.lru_cache(maxsize=5120)
```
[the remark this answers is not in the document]
Uxio0: Arbitrary. I decided it should cover enough tokens and also it's only storing a string and a number

gnosis/eth/oracles/uniswap_v3.py
Uxio0: It shouldn't. Our current implementation for Uniswap V2 relies on the router and it only returns factory for the V2. But in any case we will found out with our integration tests 😉

fmrsabino (approved):

gnosis/eth/tests/oracles/test_uniswap_v3.py
```diff
@@ -0,0 +1,68 @@
...
+        price = uniswap_v3_oracle.get_price(
+            usdt_token_mainnet_address, usdc_token_mainnet_address
+        )
+        self.assertAlmostEqual(price, 1.0, delta=0.5)
+
+        # Test with 2 stablecoins with different decimals
+        price = uniswap_v3_oracle.get_price(
+            dai_token_mainnet_address, usdc_token_mainnet_address
+        )
+        self.assertAlmostEqual(price, 1.0, delta=0.5)
```
moisses89, line 48: High delta for a stablecoin I hope this don't happens! ;)

Uxio0: Yeah, just in case 😉

PR #353, opened 2022-09-09
Add Cowswap oracle

Uxio0:
- Related to https://github.com/safe-global/safe-transaction-service/issues/645

### DIFF
--- a/gnosis/eth/oracles/cowswap.py
+++ b/gnosis/eth/oracles/cowswap.py
@@ -52,17 +52,23 @@ class CowswapOracle(PriceOracle):
             return 1.0
 
         token_1_decimals = get_decimals(token_address_1, self.ethereum_client)
-        result = self.api.get_estimated_amount(
-            token_address_1, token_address_2, OrderKind.SELL, 10**token_1_decimals
-        )
-        if "errorType" in result:
-            error_message = (
-                f"Cannot get price from CowSwap {result} "
-                f"for token-1={token_address_1} to token-2={token_address_2}"
+        try:
+            result = self.api.get_estimated_amount(
+                token_address_1, token_address_2, OrderKind.SELL, 10**token_1_decimals
             )
-            logger.warning(error_message)
-            raise CannotGetPriceFromOracle(error_message)
-        else:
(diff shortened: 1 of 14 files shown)


gnosis/eth/oracles/cowswap.py
```diff
@@ -0,0 +1,74 @@
...
+
+        token_1_decimals = get_decimals(token_address_1, self.ethereum_client)
+        try:
+            result = self.api.get_estimated_amount(
+                token_address_1, token_address_2, OrderKind.SELL, 10**token_1_decimals
+            )
+            if "amount" in result:
+                # Decimals needs to be adjusted
+                token_2_decimals = get_decimals(token_address_2, self.ethereum_client)
+                return float(result["amount"]) / 10**token_2_decimals
```
fmrsabino, line 62: Should we use `fractions` here since we are dealing with financial decimal values?

gnosis/eth/tests/oracles/test_cowswap.py
```diff
@@ -0,0 +1,85 @@
...
+    usdc_token_mainnet_address,
+    usdt_token_mainnet_address,
+    weth_token_mainnet_address,
+)
+from ..utils import just_test_if_mainnet_node
+
+
+class TestCowswapOracle(EthereumTestCaseMixin, TestCase):
+    def test_get_price(self):
+        mainnet_node = just_test_if_mainnet_node()
```
fmrsabino, line 22: OOC: is this testing the real node (E2E)?

fmrsabino (approved):

gnosis/eth/oracles/cowswap.py
Uxio0: That's a good idea, but I wouldn't do it as it would break oracle interface for all the implemented oracles. At some point we can refactor all the oracles to return `Fraction`

gnosis/eth/tests/oracles/test_cowswap.py
Uxio0: That's right, we are doing E2E for all oracle providers

PR #356, opened 2022-09-20
Add method to check if Oracle is available for network

Uxio0:
- This way it can be queried and no knowledge about the Oracle is required from the user

### DIFF
--- a/gnosis/eth/oracles/oracles.py
+++ b/gnosis/eth/oracles/oracles.py
@@ -42,31 +42,45 @@ class UnderlyingToken:
     quantity: int
 
 
-class PriceOracle(ABC):
+class BaseOracle(ABC):
+    @classmethod
+    @abstractmethod
+    def is_available(
+        cls,
+        ethereum_client: EthereumClient,
+    ) -> bool:
+        """
+        :param ethereum_client:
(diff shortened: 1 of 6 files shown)


gnosis/eth/oracles/oracles.py
```diff
@@ -231,6 +256,22 @@ def __init__(
...
+    ) -> bool:
+        """
+        :param ethereum_client:
+        :return: `True` if Oracle is available for the EthereumClient provided, `False` otherwise
+        """
+        return ethereum_client.is_contract(
+            cls.ROUTER_ADDRESSES.get(
+                ethereum_client.get_network(),
+                cls.ROUTER_ADDRESSES[EthereumNetwork.MAINNET],
+            )
```
fmrsabino, line 272: Why do we default to `Mainnet` in for `UniswapV2Oracle` but not for `UniswapOracle`?

fmrsabino (approved):

gnosis/eth/oracles/oracles.py
Uxio0: Because `Uniswap` used deterministic deployment for `V2` but not for `V1`

PR #381, opened 2022-10-24
Feat: Add Superfluid Oracle token adapter 

brymut:
As a continuation from here: https://github.com/safe-global/safe-transaction-service/issues/772#issuecomment-1087463359 

This work contributes towards adding a price data for Superfluid Super Tokens with the price data of their underlying tokens.
This PR adds the Superfluid Oracle token adapter and related test.

Related Issue: https://github.com/safe-global/safe-transaction-service/issues/772 

Original issue: https://github.com/superfluid-finance/protocol-monorepo/issues/732

PR #388, opened 2022-10-28
Fix superfluid oracle

Uxio0:
- `is_available` implementation was not ok
- Related to #381


Uxio0:
@brymut I need to fix this first

PR #394, opened 2022-11-08
Stop using eip712_structs

Uxio0:
- Copy library from https://github.com/jvinet/eip712 and do some adjustments (removing dependencies, adding `typing` hints, rewriting some code, adding new methods)
- EIP712 structs is not updated and it doesn't support Dictionary inputs (classes need to be declared so it's not easy to use with input data)

### DIFF
--- a/gnosis/eth/eip712/__init__.py
+++ b/gnosis/eth/eip712/__init__.py
@@ -1,0 +1,178 @@
+"""
+Based on https://github.com/jvinet/eip712, adjustments by https://github.com/uxio0
+
+Routines for EIP712 encoding and signing.
+
+Copyright (C) 2022 Judd Vinet <jvinet@zeroflux.org>
+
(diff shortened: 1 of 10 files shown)


gnosis/eth/eip712/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+SOFTWARE.
+"""
+import re
+from typing import Any, Dict, List, Union
+
+from eth_abi import encode_abi
+from eth_account import Account
+from eth_typing import Hash32, HexStr
+
+from ..utils import fast_keccak
```
fmrsabino, line 33: This changed right (from the original)?

I think we should mention this.

fmrsabino (approved):
LGTM. I would just add a description on the changes either at the top of the file or PR.

For example `eip712_encode_hash` seems to be a new function and the usage of `fast_keccak`.

PR #399, opened 2022-11-16
Overwrite value to string for keccak256field

moisses89:
This PR overwrite value_to_string function of keccak256field because the function of BinaryField code/decode using base64 and this not works for keccak256field. 
 
- [Overwrite value to string for keccak256field](https://github.com/safe-global/safe-eth-py/commit/b3f175aa0542755165a81e806eb10a844cb2f0ea)

### DIFF
--- a/gnosis/eth/django/tests/test_models.py
+++ b/gnosis/eth/django/tests/test_models.py
@@ -148,7 +148,7 @@ class TestModels(TestCase):
             with transaction.atomic():
                 Keccak256Hash.objects.create(value=value_hex_invalid)
 
-    def test_serialize_Keccak256Field_to_json(self):
+    def test_serialize_keccak256_field_to_json(self):
         hexvalue: str = (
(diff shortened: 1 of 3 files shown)


gnosis/eth/django/tests/test_models.py
```diff
@@ -146,3 +147,40 @@ def test_keccak256_field(self):
         ):
             with transaction.atomic():
                 Keccak256Hash.objects.create(value=value_hex_invalid)
+
+    def test_serialize_Keccak256Field_to_json(self):
```
Uxio0, line 151: Use snake case, even if you are referring a class name, same with the other test methods 😉

moisses89: Done [here](https://github.com/safe-global/safe-eth-py/pull/399/commits/cc9303c3b3161ca0fffa287661d764d3e33b334c)

Uxio0 (approved):

PR #406, opened 2022-12-05
Add Arbitrum Goerli 421613

gzeoneth:


moisses89 (changes requested):
Could you add the blockscout explorer [here](https://github.com/safe-global/safe-eth-py/blob/master/gnosis/eth/clients/blockscout_client.py#L20)?
There is one blockexplorer listed https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-421613.json

Uxio0 (approved):

moisses89 (approved):
Blockexplorer is not mandatory, only a suggestion :)

PR #416, opened 2022-12-16
Update web3 to v6

Uxio0:
It requires a major refactor and is required and the first step for updating to Python v3.11:

- Update eth-abi calls (eth-abi previous version was defective, so it was our contract signature encoding/decoding)
- Use new exception types
- Use new tracing module from web3 and update our decoder to match web3 one when using batch calls (I would like to use web3 one but I was not able to)
- Update mocks as they were lacking some information for `reward` traces due to a defective decoding
- Refactor tests
- Use snake_case web3 methods instead of camelcase


fmrsabino (approved):
LGTM 👍

Uxio0:
> LGTM 👍

We still need to wait for some libraries for full Python 3.11 support 😉

rollue:
3.11 test that is breaking may have something to do with this
https://github.com/ethereum/web3.py/issues/2704

Uxio0:
We are waiting for a web3 release with Python 3.11 support

Uxio0:
Pending on https://github.com/ethereum/web3.py/issues/2830

rollue:
this seems to be closed on web3.py?

Uxio0:
> This issue seems to be merged and closed on web3.py

Sure, but tracing API was removed so we need to wait for a release: https://github.com/ethereum/web3.py/pull/2851#event-8858345074

gnosis/eth/tests/test_ethereum_client.py
```diff
@@ -765,7 +765,7 @@ def test_trace_filter(self):
             ValueError, "The method trace_filter does not exist/is not available"
         ):
             self.ethereum_client.parity.trace_filter(
```
moisses89, line 767: Maybe we should think in other name for `parity`, tracing sound like web3 methods what do you think?

gnosis/eth/tests/test_ethereum_client.py
```diff
@@ -765,7 +765,7 @@ def test_trace_filter(self):
...
-                from_address=Account.create().address
+                to_address=[Account.create().address]
```
moisses89, line 768: I didn't understand this change, it was filtered first for from_address and now for to_address. It is correct?

gnosis/eth/ethereum_client.py
```diff
@@ -1064,16 +1074,8 @@ def get_next_traces(
...
-            return self._decode_traces(
-                self.slow_w3.parity.trace_block(block_identifier)
-            )
-        except ParityTraceDecodeException as exc:
-            logger.warning("Problem decoding trace: %s - Retrying", exc)
-            return self._decode_traces(
-                self.slow_w3.parity.trace_block(block_identifier)
-            )
+    def trace_block(self, block_identifier: BlockIdentifier) -> List[BlockTrace]:
+        return self.slow_w3.tracing.trace_block(block_identifier)
```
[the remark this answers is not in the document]
Uxio0: Web3 is now decoding the traces, so we only have to decode them if using raw HTTP calls. The good solution would be to use their formatters for our methods, but it requires a little refactor: https://github.com/ethereum/web3.py/issues/2914#issuecomment-1518018273

gnosis/eth/tests/test_ethereum_client.py
Uxio0: The method is supposed to wait for a list of addresses, we had it wrong: https://openethereum.github.io/JSONRPC-trace-module#trace_filter

gnosis/eth/tests/test_ethereum_client.py
Uxio0: Sure, I fogot to rename it 😉

moisses89 (approved):
Awesome work!

PR #422, opened 2023-01-02
Fix wrong domain separator for contract versions below 1.3.0

### LINKED ISSUE #421: DomainSeparator is wrong calculated for version < 1.3.0
**Describe the bug**
For versions < v1.3.0 the chainId is not included in the domainseparator.
`domainSeparator = keccak256(abi.encode(DOMAIN_SEPARATOR_TYPEHASH, this));`

**Expected behavior**
DomainSeparator function should return correct hash for old versions.

moisses89:
Closes #421 
- Add `domainSeparator`  calculation for contract version 1.1.1
- Add test eip1271 signing for v1.1.1 `

### DIFF
--- a/gnosis/safe/safe.py
+++ b/gnosis/safe/safe.py
@@ -88,8 +88,11 @@ class Safe:
         0x4A204F620C8C5CCDCA3FD54D003BADD85BA500436A431F0CBDA4F558C93C34C8
     )
     # keccak256("EIP712Domain(uint256 chainId,address verifyingContract)");
-    DOMAIN_TYPEHASH = bytes.fromhex(
+    DOMAIN_TYPEHASH_V1_3_0 = bytes.fromhex(
         "47e79534a245952e8b16893a336b85a3d9ea9fa8c573f3d803afb92a79469218"
+    )
+    DOMAIN_TYPEHASH_V1_1_1 = bytes.fromhex(
+        "035aff83d86937d35b32e04f0ddc6ff469290eef2f1b692d8a815c89404d4749"
     )
     # keccak256("SafeMessage(bytes message)");
     SAFE_MESSAGE_TYPEHASH = bytes.fromhex(
@@ -519,16 +522,28 @@ class Safe:
 
     @cached_property
     def domain_separator(self):
-        return fast_keccak(
-            encode_abi(
-                ["bytes32", "uint256", "address"],
-                [
-                    self.DOMAIN_TYPEHASH,
-                    self.ethereum_client.get_chain_id(),
-                    self.address,
-                ],
+        version = self.contract.functions.VERSION().call()
+        if version == "1.3.0":
+            return fast_keccak(
+                encode_abi(
+                    ["bytes32", "uint256", "address"],
+                    [
+                        self.DOMAIN_TYPEHASH_V1_3_0,
+                        self.ethereum_client.get_chain_id(),
+                        self.address,
+                    ],
+                )
             )
-        )
+        else:
+            return fast_keccak(
+                encode_abi(
+                    ["bytes32", "address"],
+                    [
+                        self.DOMAIN_TYPEHASH_V1_1_1,
+                        self.address,
+                    ],
+                )
+            )
 
     def check_funds_for_tx_gas(
         self, safe_tx_gas: int, base_gas: int, gas_price: int, gas_token: str
--- a/gnosis/safe/safe.py
+++ b/gnosis/safe/safe.py
@@ -524,7 +524,7 @@ class Safe:
 
     @cached_property
     def domain_separator(self):
-        version = self.retrieve_version
+        version = self.retrieve_version()
         if version == "1.3.0":
             return fast_keccak(
                 encode_abi(
@@ -1068,7 +1068,6 @@ class Safe:
             block_identifier=block_identifier
         )
 
-    @cached_property
     def retrieve_version(
         self, block_identifier: Optional[BlockIdentifier] = "latest"
     ) -> str:
@@ -1111,7 +1110,7 @@ class Safe:
 
         if safe_nonce is None:
             safe_nonce = self.retrieve_nonce()
-        safe_version = safe_version or self.retrieve_version
+        safe_version = safe_version or self.retrieve_version()
         return SafeTx(
             self.ethereum_client,
             self.address,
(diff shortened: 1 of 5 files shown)


gnosis/safe/safe.py
```diff
@@ -519,16 +522,28 @@ def contract(self) -> Contract:
...
     def domain_separator(self):
-        return fast_keccak(
-            encode_abi(
-                ["bytes32", "uint256", "address"],
-                [
-                    self.DOMAIN_TYPEHASH,
-                    self.ethereum_client.get_chain_id(),
-                    self.address,
-                ],
+        version = self.contract.functions.VERSION().call()
```
moisses89, line 525: I am following the implementation of `domain_separator` which is computed instead of calling the contract.
I'm not sure why, but maybe this was implemented at this way to avoid asking the blockchain one more time.
If this were the reason, then this implementation should be changed to create a `@cached_property` of the contract version or perhaps call `domain_separator` directly to the blockchain instead of requesting the version once more.

gnosis/safe/safe.py
```diff
@@ -88,9 +88,12 @@ class Safe:
         0x4A204F620C8C5CCDCA3FD54D003BADD85BA500436A431F0CBDA4F558C93C34C8
     )
     # keccak256("EIP712Domain(uint256 chainId,address verifyingContract)");
-    DOMAIN_TYPEHASH = bytes.fromhex(
+    DOMAIN_TYPEHASH_V1_3_0 = bytes.fromhex(
         "47e79534a245952e8b16893a336b85a3d9ea9fa8c573f3d803afb92a79469218"
     )
+    DOMAIN_TYPEHASH_V1_1_1 = bytes.fromhex(
+        "035aff83d86937d35b32e04f0ddc6ff469290eef2f1b692d8a815c89404d4749"
```
fmrsabino, line 95: I'd add a similar comment on what resulted in this hash value:

```python
# keccak256("EIP712Domain(address verifyingContract)")
```

gnosis/safe/safe.py
fmrsabino: I'd say that the library was just assuming 1.3.0 for the domain separator. Therefore, it wouldn't perform any RPC calls given that assumption.

fmrsabino, line 525: What about calling `retrieve_version()` here?

I think we can make it a `cached_property` in a separate PR since it affects other methods.

gnosis/safe/safe.py
```diff
@@ -519,16 +522,28 @@ def contract(self) -> Contract:
...
+                encode_abi(
+                    ["bytes32", "uint256", "address"],
+                    [
+                        self.DOMAIN_TYPEHASH_V1_3_0,
+                        self.ethereum_client.get_chain_id(),
+                        self.address,
+                    ],
+                )
+            )
+        else:
```
fmrsabino, line 537: I think we should check the version here again and throw an error if no version was matched.

Else we are implicitly falling back to version 1 and ignoring a possible error.

fmrsabino (changes requested):

gnosis/safe/safe.py
moisses89: Done https://github.com/safe-global/safe-eth-py/pull/422/commits/f1ac04392efe6c31df0dc636a931ade86c92155e

gnosis/safe/safe.py
moisses89: Done https://github.com/safe-global/safe-eth-py/pull/422/commits/f1ac04392efe6c31df0dc636a931ade86c92155e

gnosis/safe/safe.py
moisses89: Done https://github.com/safe-global/safe-eth-py/pull/422/commits/f1ac04392efe6c31df0dc636a931ade86c92155e 
I added an Exception to indicate a version error.

gnosis/safe/tests/test_signatures.py
```diff
@@ -11,6 +11,9 @@
 
 
 class TestSafeSignature(SafeTestCaseMixin, TestCase):
+    EIP1271_MAGIC_VALUE = "20c13b0b"
+    UPDATED_MAGIC_VALUE = "1626ba7e"
```
fmrsabino, line 15: nitpick: this is still related with EIP1271. I would call it `EIP1271_MAGIC_VALUE_UPDATED` which I think better represents an update on original value.

gnosis/safe/safe.py
```diff
@@ -1047,6 +1068,7 @@ def retrieve_threshold(
             block_identifier=block_identifier
         )
 
+    @cached_property
```
fmrsabino, line 1071: nitpick: since the consumers of this function will see it as a property I'd remove the `retrieve` word.

i.e. `contract.version` looks more like a property than `contract.retrieve_version`. The fact that it actually retrieves something and caches it should be hidden to whoever calls this.

moisses89: Makes sense but maybe shouldn't be done here, what do you think if I revert the cached_property and create an issue to add the property version to the contract class in a different PR?

fmrsabino: > Makes sense but maybe shouldn't be done here, what do you think if I revert the cached_property and create an issue to add the property version to the contract class in a different PR? 

I'm fine with that too. 

> Also I'm concerned about what would happen if the contract is updated to new version and this property is cached.

This should be an edge case which can be worked around by creating a new instance of the class. Even though it can happen we should compare this edge case to the possible reduction in RPC calls. But we can discuss this separately if you are going to revert this change 👍

gnosis/safe/tests/test_signatures.py
moisses89: Makes sense, https://github.com/safe-global/safe-eth-py/pull/422/commits/04fdade907f04822b203b1dac9d7e78696c5aed1

gnosis/safe/safe.py
moisses89: Reverted here https://github.com/safe-global/safe-eth-py/pull/422/commits/f7cee19541ea4af81177fca1eb841114e02e0e5f

fmrsabino (approved):
🚀 🚀 🚀

gnosis/safe/safe.py
Uxio0: Agree on this:

`i.e. contract.version looks more like a property than contract.retrieve_version. The fact that it actually retrieves something and caches it should be hidden to whoever calls this.`

The only thing pulling me out from caching version is that it can change and break existing services

PR #427, opened 2023-01-09
Add EIP712 support for bytes and uint formatted as string

### LINKED ISSUE #424: eip712_encode_hash encode string bytes and string uint256
# What is needed?
## String bytes feature
I would like that `eip712_encode_hash` encode string `bytes` also like the following example:
`"Message": [
            {"name": "oneByte", "type": "bytes1"}
        ]
   "message": {"oneByte": "0x01"}`

Currently this is not possible because `eip712_encode_hash` is using `eth_abi` that expects python bytes.

## String uint256 feature
I would like that `eip712_encode_hash` encode string `uint` also like the following example:
`"Message": [
            {"name": "stringNum", "type": "uint256"}
        ]
   "message": {"stringNum": "1234"}`
   
# Possible solution
Check in [_encode_field(name, typ, value)](https://github.com/safe-global/safe-eth-py/blob/master/gnosis/eth/eip712/__init__.py#L36) if is bytes type the value is an instance of `str` and in that case convert to bytes `bytes.fromhex(string)`.
Similar solution for `uint` but converting it from string to python int.
# Related issues

moisses89:
Closes #424 
# Description
`eip712_encode_hash` is using `eth-abi`  that doesn't support the type `bytes` and `uint` formatted as string. 
Type `bytes` and `bytesN` on `eth-abi` only accept python bytes, this was fixed including a conversion for bytes formatted on string to python bytes.
Type `uint` only accept Numbers, this was fixed converting from string to python Integer.
# Additional changes
- Add test for string uint
- Add test for string bytes

### DIFF
--- a/gnosis/eth/eip712/__init__.py
+++ b/gnosis/eth/eip712/__init__.py
@@ -57,6 +57,14 @@ def encode_data(primary_type: str, data, types):
         if value is None:
             raise Exception(f"Missing value for field {name} of type {type}")
 
+        # Accept string bytes
+        if "bytes" in typ and isinstance(value, str):
+            value = bytes.fromhex(value.replace("0x", ""))
+
+        # Accept string uint
+        if "uint" in typ and isinstance(value, str):
+            value = int(value)
+
         if typ == "bytes":
             return ["bytes32", fast_keccak(value)]
 
--- a/gnosis/eth/tests/eip712/test_eip712.py
+++ b/gnosis/eth/tests/eip712/test_eip712.py
@@ -118,7 +118,6 @@ class TestEIP712(TestCase):
                 {"name": "maxByte", "type": "bytes32"},
             ],
         }
-        # test string uint (chainId)
         payload = {
             "types": types_with_bytes,
             "primaryType": "Message",
--- a/gnosis/eth/tests/eip712/test_eip712.py
+++ b/gnosis/eth/tests/eip712/test_eip712.py
@@ -4,6 +4,43 @@ from gnosis.eth.eip712 import eip712_encode_hash
 
 
 class TestEIP712(TestCase):
+    address = "0x8e12f01dae5fe7f1122dc42f2cb084f2f9e8aa03"
+    types = {
+        "EIP712Domain": [
+            {"name": "name", "type": "string"},
+            {"name": "version", "type": "string"},
+            {"name": "chainId", "type": "uint256"},
+            {"name": "verifyingContract", "type": "address"},
+        ],
+        "Mailbox": [
+            {"name": "owner", "type": "address"},
+            {"name": "messages", "type": "Message[]"},
+        ],
+        "Message": [
+            {"name": "sender", "type": "address"},
+            {"name": "subject", "type": "string"},
+            {"name": "isSpam", "type": "bool"},
+            {"name": "body", "type": "string"},
+        ],
+    }
+
+    msgs = [
+        {
+            "sender": address,
+            "subject": "Hello World",
+            "body": "The sparrow flies at midnight.",
+            "isSpam": False,
+        },
+        {
+            "sender": address,
+            "subject": "You may have already Won! :dumb-emoji:",
+            "body": "Click here for sweepstakes!",
+            "isSpam": True,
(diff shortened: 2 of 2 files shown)


gnosis/eth/tests/eip712/test_eip712.py
```diff
@@ -33,59 +68,70 @@ def test_eip712_encode_hash(self):
...
+        types_with_bytes = {
+            "EIP712Domain": [
+                {"name": "name", "type": "string"},
+            ],
+            "Message": [
+                {"name": "oneByte", "type": "bytes1"},
+                {"name": "maxByte", "type": "bytes32"},
+            ],
+        }
+        # test string uint (chainId)
```
fmrsabino, line 121: nitpick: this comment does not apply to this test right?

fmrsabino (approved):
Nice 👏 🚀

gnosis/eth/tests/eip712/test_eip712.py
moisses89: Yes, copy pasted from above, done https://github.com/safe-global/safe-eth-py/pull/427/commits/fa4a21f60b881a33fd2c0b8a72aa3c5c3c74b188

fmrsabino (approved):

gnosis/eth/eip712/__init__.py
```diff
@@ -57,6 +57,14 @@ def _encode_field(name, typ, value):
         if value is None:
             raise Exception(f"Missing value for field {name} of type {type}")
 
+        # Accept string bytes
+        if "bytes" in typ and isinstance(value, str):
+            value = bytes.fromhex(value.replace("0x", ""))
```
Uxio0, line 62: `value = HexBytes(value)`

gnosis/eth/eip712/__init__.py
```diff
@@ -57,6 +57,14 @@ def _encode_field(name, typ, value):
...
+
+        # Accept string uint
```
Uxio0, line 64: I think we should support any kind of integer, not only `uint`. This is a very misleading behaviour:

`if "int" in typ`

gnosis/eth/tests/eip712/test_eip712.py
```diff
@@ -4,6 +4,43 @@
 
 
 class TestEIP712(TestCase):
+    address = "0x8e12f01dae5fe7f1122dc42f2cb084f2f9e8aa03"
```
Uxio0, line 7: We should use checksummed addresses unless we are testing not checksummed addresses:

`0x8e12f01DAE5FE7f1122Dc42f2cB084F2f9E8aA03`

gnosis/eth/tests/eip712/test_eip712.py
```diff
@@ -33,59 +68,69 @@ def test_eip712_encode_hash(self):
...
+            },
+            "message": {
+                "oneByte": "0x01",
+                "maxByte": "0x6214da6089b8d8aaa6e6268977746aa0af19fd1ef5d56e225bb3390a697c3ec1",
+            },
+        }
+
+        self.assertEqual(
+            eip712_encode_hash(payload).hex(),
+            "2950cf06416c6c20059f24a965e3baf51a24f4ef49a1e7b1a47ee13ee08cde1f",
```
Uxio0, line 135: Why do you use `hex()` here? I think we should be consistent, either using `hex()` everywhere (it looks better and more readable IMHO) or using binary representation

Uxio0 (commented):
@moisses89 I think these things should be addressed in a new PR

gnosis/eth/tests/eip712/test_eip712.py
moisses89: > @moisses89 I think these things should be addressed in a new PR
Your suggestions sounds good for me, then I add the following PR https://github.com/safe-global/safe-eth-py/pull/437
Regarding if use hex() or binary representation for test, I'd prefer hex() because as you mention is more readable.

Uxio0: Sounds good, thanks

PR #428, opened 2023-01-09
Bump postgres version

moisses89:
The current django version don't support `postgresql-10` : 
`FAILED gnosis/safe/tests/api/test_transaction_service_api.py::TestTransactionServiceAPI::test_get_balances - django.db.utils.NotSupportedError: PostgreSQL 11 or later is required (found 10.23).`
The version 13 was chosen to be aligned with our github actions https://github.com/safe-global/safe-eth-py/blob/master/.github/workflows/python.yml#L41

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: '3.5'
 
 services:
   db:
-    image: postgres:10-alpine
+    image: postgres:15-alpine
     ports:
       - "5432:5432"
     environment:


docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3.5'
 
 services:
   db:
-    image: postgres:10-alpine
+    image: postgres:13-alpine
```
fmrsabino, line 5: Why not use the latest stable? `postgres:15-alpine`

moisses89: I only tried with the same version that is for github actions, I'm happy using the last version of postgres for locally tests.  I just updated to `postgres:15-alpine`.

moisses89: I tested it locally also.

fmrsabino (approved):

docker-compose.yml
Uxio0: We were using `postgres:13` just because is the last version supported by AWS, but `postgres:15` should do it

PR #432, opened 2023-01-12
Set version 4.9.1

moisses89:
# Commit included 
[Fix wrong domain separator for contract version 1.1.1 ](https://github.com/safe-global/safe-eth-py/commit/c0001208cb66c03ccae30b627b17c78c34bb5fe0)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = safe-eth-py
-version = 4.9.0
+version = 4.9.1
 description = Safe Ecosystem Foundation utilities for Ethereum projects
 long_description = file: README.rst
 long_description_content_type = text/x-rst; charset=UTF-8


docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3.5'
 
 services:
   db:
-    image: postgres:10-alpine
```
fmrsabino, line 5: We should just update the version in this PR.

fmrsabino (changes requested):

docker-compose.yml
moisses89: Ups, yes :)

fmrsabino (approved):

docker-compose.yml
moisses89: Fixed, removed the commit upload by mistake.

PR #438, opened 2023-01-16
Use onchain domainSeparator instead of local calculation

Uxio0:
- Versions different from 1.3.0 or 1.1.1 were not supported
- Related to #421

### DIFF
--- a/gnosis/safe/safe.py
+++ b/gnosis/safe/safe.py
@@ -6,6 +6,7 @@ from logging import getLogger
 from typing import Callable, List, NamedTuple, Optional, Union
 
 from eth_abi import encode_abi
+from eth_abi.exceptions import DecodingError
 from eth_abi.packed import encode_abi_packed
 from eth_account import Account
 from eth_account.signers.local import LocalAccount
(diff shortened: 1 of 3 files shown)


gnosis/safe/safe.py
```diff
@@ -115,6 +108,26 @@ def __init__(self, address: ChecksumAddress, ethereum_client: EthereumClient):
...
         return f"Safe={self.address}"
 
+    @cached_property
+    def contract(self) -> Contract:
+        v_1_3_0_contract = get_safe_V1_3_0_contract(self.w3, address=self.address)
+        version = v_1_3_0_contract.functions.VERSION().call()
+        if version == "1.3.0":
+            return v_1_3_0_contract
+        else:
+            return get_safe_V1_1_1_contract(self.w3, address=self.address)
```
moisses89, line 118: One question about this, why we return the contract 1.1.1 when for example the contract version is 1.2.0?

Uxio0: Because they share "almost" the same ABI

moisses89 (approved):

Issue #440, opened 2023-01-17
CowSwap oracle doesn't work on gnosis-chain


moisses89:
**Describe the bug**
CowSwap oracle currently doesn't work on `safe-eth-py` because the endpoint to get the price was deprecated and removed. 
https://github.com/cowprotocol/services/pull/1052/files#diff-ea5f6887f8360abaf702f3dcece07c3b2e87022458254f48bb2fb66adf15b95fL270

**To Reproduce**
Try to get any price using cowswap oracle. 

**Expected behavior**
Return the price

**Possible solution**
There is other endpoint that can return the native price:
`https://api.cow.fi/xdai/api/v1/token/<token-address>/native_price`
For example:
[https://api.cow.fi/xdai/api/v1/token/0x177127622c4A00F3d409B75571e12cB3c8973d3c/native_price](https://api.cow.fi/xdai/api/v1/token/0x177127622c4A00F3d409B75571e12cB3c8973d3c/native_price)

PR #445, opened 2023-01-19
Use new quote API for cowswap oracle

Uxio0:
- Markets API was deprecated, quote API is used now
- Closes https://github.com/safe-global/safe-eth-py/issues/440

### DIFF
--- a/gnosis/protocol/gnosis_protocol_api.py
+++ b/gnosis/protocol/gnosis_protocol_api.py
@@ -10,6 +10,7 @@ from hexbytes import HexBytes
 from gnosis.eth import EthereumNetwork, EthereumNetworkNotSupported
 from gnosis.eth.eip712 import eip712_encode_hash
 
+from ..eth.constants import NULL_ADDRESS
 from .order import Order, OrderKind
 
 
@@ -27,8 +28,8 @@ class TradeResponse(TypedDict):
 
 
 class AmountResponse(TypedDict):
-    amount: str
-    token: AnyAddress
+    sellAmount: int
+    buyAmount: int
 
 
 class ErrorResponse(TypedDict):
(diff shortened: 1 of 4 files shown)


moisses89:
This is going to close https://github.com/safe-global/safe-eth-py/issues/440

gnosis/protocol/gnosis_protocol_api.py
```diff
@@ -96,10 +99,20 @@ def get_fee(self, order: Order, from_address: ChecksumAddress) -> int:
...
 
+    def get_fee(
+        self, order: Order, from_address: ChecksumAddress
+    ) -> Union[int, ErrorResponse]:
+        quote = self.get_quote(order, from_address)
+
+        if "quote" in quote:
+            return int(quote["quote"]["feeAmount"])
+        else:
+            return quote
```
moisses89, line 114: here we are returning an error?

gnosis/protocol/gnosis_protocol_api.py
```diff
@@ -193,14 +206,36 @@ def get_estimated_amount(
...
-        url = self.base_url + f"markets/{base_token}-{quote_token}/{kind.name}/{amount}"
-        r = self.http_session.get(url)
-        if r.ok:
-            return AmountResponse(r.json())
+        order = Order(
+            sellToken=base_token,
+            buyToken=quote_token,
+            receiver=NULL_ADDRESS,
+            sellAmount=amount_wei * 10,
+            buyAmount=0,
```
moisses89, line 224: buyAmount=0 are you sure about this? here if the response from cowswap return 0 then the result is going to be 0 always. 
https://github.com/safe-global/safe-eth-py/pull/445/files#diff-800f25d9436120b4a629a7266b15f2b500b0a01ec5d7ca55c8509a6098f0be1cR63

gnosis/protocol/tests/test_gnosis_protocol_api.py
```diff
@@ -63,7 +63,7 @@ def test_get_estimated_amount(self):
             OrderKind.SELL,
             int(1e18),
         )
-        amount = int(response["amount"]) / 1e18
+        amount = float(response["buyAmount"]) / response["sellAmount"]
```
moisses89, line 66: the same as above if buyAmount is 0.

gnosis/protocol/gnosis_protocol_api.py
Uxio0: Yes, we should use exceptions, I don't know why I did it this way (like Golang), but maybe it's too much to refactor everything now

gnosis/protocol/gnosis_protocol_api.py
Uxio0: You are right, fixed

moisses89 (approved):
Looks good to me!

PR #457, opened 2023-02-06
Don't log price oracle failures

Uxio0:
It's not relevant if a price oracle cannot retrieve the price, as there's already an exception returned for that

- Update price oracle logs from `WARNING` to `DEBUG`

### DIFF
--- a/gnosis/eth/oracles/utils.py
+++ b/gnosis/eth/oracles/utils.py
@@ -32,6 +32,6 @@ def get_decimals(
         BadFunctionCallOutput,
         DecodingError,
     ) as e:
-        error_message = f"Cannot get decimals for token={token_address}"
-        logger.warning(error_message)
-        raise CannotGetPriceFromOracle(error_message) from e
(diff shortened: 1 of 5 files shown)


gnosis/eth/oracles/utils.py
```diff
@@ -32,6 +32,6 @@ def get_decimals(
         BadFunctionCallOutput,
         DecodingError,
     ) as e:
-        error_message = f"Cannot get decimals for token={token_address}"
-        logger.warning(error_message)
-        raise CannotGetPriceFromOracle(error_message) from e
+        message = f"Cannot get decimals for token={token_address}"
```
moisses89, line 35: Is this related with the price? I mean, ERC20 should have decimals method.

Uxio0: Yes, but when we are calculating the price we use anything that looks like an ERC20, so I wouldn't log this kind of errors. This method is only used from `oracles`

moisses89 (approved):

PR #463, opened 2023-02-15
Add more E2E `trace_block` tests for Erigon

Uxio0:
- We were missing some E2E for previously problematic blocks


moisses89 (approved):
Looks good to me, could you add in the description what is adding this block or what test was missing?

Uxio0:
> Looks good to me, could you add in the description what is adding this block or what test was missing?

Oh, I see with code formatting is not clear, I added one more block that it was problematic before `15630274`

PR #473, opened 2023-03-07
Add Safe contract addresses

Uxio0:
- Move Safe contract addresses from https://github.com/safe-global/safe-transaction-service to here, as it makes more sense and will allow libraries depending on `safe-eth-py` to use them


moisses89 (approved):
Looks good to me!

PR #477, opened 2023-03-17
Add EIP155 support for some methods

### LINKED ISSUE #476: only replay-protected (EIP-155) transactions allowed over RPC
**Describe the bug**
The current ethereum_client, when performing send_eth_to, some chains cannot support it well, and returns an error "only replay-protected (EIP-155) transactions allowed over RPC"

**To Reproduce**
1. when blockchain is avax or some other chain,it will display the error

**Expected behavior**
Line 1939 is as below:

> tx: TxParams = {
>             "from": account.address,
>             "to": to,
>             "value": value,
>             "gas": gas or Wei(self.estimate_gas(to, account.address, value)),
>             "gasPrice": Wei(gas_price),
>         }
> 
> 

and it can modify as:

> tx: TxParams = {
>             "from": account.address,
>             "to": to,
>             "value": value,
>            "chainId: self.get_chain_id(),
>             "gas": gas or Wei(self.estimate_gas(to, account.address, value)),
>             "gasPrice": Wei(gas_price),
>         }

Uxio0:
- Some methods on `EthereumClient` did not support EIP155
- Closes #476

### DIFF
--- a/gnosis/eth/ethereum_client.py
+++ b/gnosis/eth/ethereum_client.py
@@ -1407,10 +1407,11 @@ class EthereumClient:
 
     def get_network(self) -> EthereumNetwork:
         """
-        Get network name based on the chainId
+        Get network name based on the chainId. This method is not cached as the method for getting the
+        `chainId` already is.
 
         :return: EthereumNetwork based on the chainId. If network is not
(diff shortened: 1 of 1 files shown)


moisses89 (approved):
Looks good to me!

PR #482, opened 2023-03-26
Add flare chain

pkakelas:
Deployed the contracts using this PR: https://github.com/safe-global/safe-singleton-factory/pull/77/files

Thanks :)


Uxio0 (changes requested):
Contracts are not matching the expected addresses

PR #501, opened 2023-04-13
Fix WETH for CowSwap Oracle (Gnosis Chain)

Uxio0:
- CowSwap oracle in Gnosis Chain is supposed to use Wrapped XDAI for getting the price, not Wrapped Ether

From now on, `Wrapped XDAI` will be used as the default collateral for `Cowswap` oracle


moisses89 (approved):
Looks good to me!

PR #512, opened 2023-04-17
Fix cowswap place_order test

Uxio0:


gnosis/protocol/tests/test_gnosis_protocol_api.py
```diff
@@ -151,10 +151,7 @@ def test_place_order(self):
...
 
         if type(order_id) is dict:
-            if order_id["errorType"] == "NoLiquidity":
-                pytest.xfail("NoLiquidity Error")
-            elif order_id["errorType"] == "InsufficientBalance":
-                pytest.xfail("InsufficientBalance")
-
-        self.assertEqual(order_id[:2], "0x")
-        self.assertEqual(len(order_id), 114)
+            pytest.xfail(order_id["errorType"])
```
moisses89, line 154: Ok, maybe better, I didn't see that the else was missing.

moisses89 (approved):

gnosis/protocol/tests/test_gnosis_protocol_api.py
Uxio0: When doing `xfail` test stops, so else should not be required 😉

Issue #519, opened 2023-04-25
Add the 1.3.0 singleton addresses deployed with safe singleton factory as supported networks


moisses89:
# What is needed?
We would like to add the 1.3.0 singleton addresses deployed with the safe singleton factory as supported networks. ([singleton addresses](https://github.com/safe-global/safe-contracts/blob/main/CHANGELOG.md#version-130-libs0))
Also it is required  that every address version 1.3.0 that wasn't deployed with the safe singleton factory and was deployed with the default singleton factory were included here. This is being solved on 1.4.0 https://github.com/safe-global/safe-contracts/issues/460

# Motivation
The motivation of this is that all available addresses on every network work with our service.

# Deployed using Safe singleton factory 
| Contract                     | Address                                    | mainnet                                                                                | goerli                                                                                     | polygon                                                                                   | optimism                                                                                       | Gnosis-chain                                                                         | arbitrum one                                                                          | bnb                                                                                   | aurora                                                                                        | avalanche                                                                           | CELO                                                                                             | Base                                                                                       |
|------------------------------|--------------------------------------------|----------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------|
[...]

PR #539, opened 2023-05-09
Add singleton addresses

moisses89:
Related with issue #519 
Safe singleton addresses added on:
- mainnet
- goerli
- polygon
- gnosis-chain
- arbitrum

Default singleton address added on: 
- optimism
- avalanche

### DIFF
--- a/gnosis/safe/addresses.py
+++ b/gnosis/safe/addresses.py
@@ -133,7 +133,16 @@ MASTER_COPIES: Dict[EthereumNetwork, List[Tuple[str, int, str]]] = {
         ("0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552", 13736914, "1.3.0"),
     ],
     EthereumNetwork.ARBITRUM_ONE: [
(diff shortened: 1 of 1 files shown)


gnosis/safe/addresses.py
```diff
@@ -103,8 +166,23 @@
         ("0x69f4D1788e39c87893C980c06EdF4b7f686e2938", 8944351, "1.3.0"),
     ],
     EthereumNetwork.AVALANCHE_C_CHAIN: [
-        ("0xfb1bffC9d739B8D520DaF37dF666da4C687191EA", 4_949_507, "1.3.0+L2"),
-        ("0x69f4D1788e39c87893C980c06EdF4b7f686e2938", 4_949_512, "1.3.0"),
+        (
+            "0x3E5c63644E683549055b9Be8653de26E0B4CD36E",
+            22_123_383,
```
moisses89, line 171: @Uxio0 I think that we should keep this format in general, maybe we can handle it in other PR, what do you think?

PR #553, opened 2023-05-23
Fix infinite loop in `retrieve_modules`

Uxio0:
- When a Safe was unitialized `retrieve_modules` could get into an infinite loop

### DIFF
--- a/gnosis/safe/safe.py
+++ b/gnosis/safe/safe.py
@@ -904,12 +904,15 @@ class Safe:
                 raise_exception=False,
             )
             modules_response, nonce, owners, threshold, version = results
-            if modules_response:
-                modules, next_module = modules_response
-            if (
-                not modules_response or next_module != SENTINEL_ADDRESS
-            ):  # < 1.1.1 or still more elements in the list
+            if not modules_response:
+                # < 1.1.1
                 modules = self.retrieve_modules()
+            else:
+                modules, next_module = modules_response
+                if modules and next_module != SENTINEL_ADDRESS:
+                    # Still more elements in the list
+                    modules = self.retrieve_modules()
+
             return SafeInfo(
                 self.address,
                 fallback_handler,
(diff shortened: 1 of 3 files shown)


moisses89 (dismissed):

gnosis/safe/safe.py
```diff
@@ -970,10 +973,12 @@ def retrieve_modules(
         self,
         pagination: Optional[int] = 50,
         block_identifier: Optional[BlockIdentifier] = "latest",
-    ) -> List[str]:
+        max_modules_to_retrieve: Optional[int] = 500,
```
moisses89, line 976: Could this limit affect us? I mean could we reach this limit in some cases?

gnosis/safe/safe.py
```diff
@@ -989,17 +994,24 @@ def retrieve_modules(
...
+
+        for _ in range(max_modules_to_retrieve // pagination):
+            # If we use a `while True` loop a custom coded Safe could get us into an infinite loop
             (modules, address) = contract.functions.getModulesPaginated(
                 address, pagination
             ).call(block_identifier=block_identifier)
 
-            all_modules.extend(modules)
-            if address == SENTINEL_ADDRESS:
+            if not modules or address in (NULL_ADDRESS, SENTINEL_ADDRESS):
```
moisses89, line 1005: Maybe could be a while loop: 
`while modules and address not in (NULL_ADDRESS, SENTINEL_ADDRESS): `

gnosis/safe/safe.py
Uxio0: I don't think a Safe could have such a number of modules

gnosis/safe/safe.py
Uxio0: If a person deploys a malicious Safe we could get into an infinite loop, that's what I want to prevent

gnosis/safe/safe.py
```diff
@@ -989,17 +994,24 @@ def retrieve_modules(
...
+                # `NULL_ADDRESS` is only seen in uninitialized Safes
                 break
-            else:
-                all_modules.append(address)
+
+            # Safes with version < 1.4.0 don't include the `address` used as pagination in the module list
+            # From 1.4.0 onwards it is included, so we check for duplicated addresses before inserting
```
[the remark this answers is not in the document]
Uxio0: Hi @mmv08 , that's what I meant. Previously the `next` address if used for pagination was not included in the list, and 1.4.0 it does. How could I make it more clear in the comment?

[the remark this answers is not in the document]
Uxio0: Makes sense, I will add it to the comment

PR #555, opened 2023-05-25
Add helper method to convert RPC calls to JSON

Uxio0:
- Very useful to compare requests, as it will sort the keys

### DIFF
--- a/gnosis/eth/tests/utils.py
+++ b/gnosis/eth/tests/utils.py
@@ -1,8 +1,12 @@
+import json
 import os
+from copy import deepcopy
+from typing import Any
 
 import pytest
 import requests
 from eth_account.signers.local import LocalAccount
+from hexbytes import HexBytes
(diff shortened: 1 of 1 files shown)


gnosis/eth/tests/utils.py
```diff
@@ -140,3 +144,35 @@ def deploy_erc20(
...
+    elif isinstance(o, set):
+        o = {bytes_to_str(element) for element in o}
+    return o
+
+
+def to_json_with_hexbytes(o: Any) -> str:
+    """
+    Convert RPC calls with nested bytes/Hexbytes to json and compare. Useful for RPC calls
+
+    :param o:
```
moisses89, line 175: what is `o` param?

Uxio0: Object, but `object` is a reserved word so I used `o`

PR #568, opened 2023-06-22
Fix concurrent HTTP connection issues

Uxio0:
- Add missing `timeout` for some http queries
- Increase connection pool for Coingecko and Cow API
- Fix typo and reduce number of retries for EthereumClient

### DIFF
--- a/gnosis/protocol/gnosis_protocol_api.py
+++ b/gnosis/protocol/gnosis_protocol_api.py
@@ -71,7 +71,7 @@ class GnosisProtocolAPI:
         """
         Prepare http session with custom pooling. See:
         https://urllib3.readthedocs.io/en/stable/advanced-usage.html
-        https://2.python-requests.org/en/latest/api/#requests.adapters.HTTPAdapter
+        https://docs.python-requests.org/en/v1.2.3/api/#requests.adapters.HTTPAdapter
         https://web3py.readthedocs.io/en/stable/providers.html#httpprovider
         """
         session = requests.Session()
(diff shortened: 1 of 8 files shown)


gnosis/protocol/gnosis_protocol_api.py
```diff
@@ -64,7 +64,25 @@ def __init__(self, ethereum_network: EthereumNetwork):
...
         self.base_url = self.API_BASE_URLS[self.network]
-        self.http_session = requests.Session()
+        self.http_session = self._prepare_http_session()
+        self.request_timeout = request_timeout
+
+    def _prepare_http_session(self) -> requests.Session:
+        """
+        Prepare http session with custom pooling. See:
+        https://urllib3.readthedocs.io/en/stable/advanced-usage.html
+        https://2.python-requests.org/en/latest/api/#requests.adapters.HTTPAdapter
```
moisses89, line 74: This link doesn't open for me due SSL issues, I think that this works better https://docs.python-requests.org/en/v1.2.3/api/#requests.adapters.HTTPAdapter

gnosis/safe/api/base_api.py
```diff
@@ -36,6 +37,7 @@ def __init__(
         if not self.base_url:
             raise EthereumNetworkNotSupported(network)
         self.http_session = requests.Session()
```
moisses89, line 39: Maybe you forgot to enable here the pool connection for this requests?

Uxio0: Not really required, but adding it

moisses89 (approved):

PR #586, opened 2023-07-12
feat: add godwoken mainnet addresses

Kuzirashi:
Based on: https://github.com/safe-global/safe-deployments/pull/147

### DIFF
--- a/gnosis/eth/clients/blockscout_client.py
+++ b/gnosis/eth/clients/blockscout_client.py
@@ -40,7 +40,7 @@ class BlockscoutClient:
         EthereumNetwork.REI_CHAIN_TESTNET: "https://scan-test.rei.network/",
         EthereumNetwork.METER_MAINNET: "https://scan.meter.io/",
         EthereumNetwork.METER_TESTNET: "https://scan-warringstakes.meter.io/",
-        EthereumNetwork.GODWOKEN_TESTNET_V1: "https://v1.betanet.gwscan.com/",
+        EthereumNetwork.GODWOKEN_TESTNET_V1: "https://v1.testnet.gwscan.com/",
         EthereumNetwork.GODWOKEN_MAINNET: "https://v1.gwscan.com/",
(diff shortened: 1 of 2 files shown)


gnosis/eth/clients/blockscout_client.py
```diff
@@ -40,7 +40,7 @@ class BlockscoutClient:
         EthereumNetwork.REI_CHAIN_TESTNET: "https://scan-test.rei.network/",
         EthereumNetwork.METER_MAINNET: "https://scan.meter.io/",
         EthereumNetwork.METER_TESTNET: "https://scan-warringstakes.meter.io/",
-        EthereumNetwork.GODWOKEN_TESTNET_V1: "https://v1.betanet.gwscan.com/",
+        EthereumNetwork.GODWOKEN_TESTNET_V1: "https://v1.testnet.gwscan.com/",
```
moisses89, line 43: Is this blockscout api compatible?

moisses89 (commented):
Could you verify the contracts in the blockexplorer?

Uxio0:
@Kuzirashi are you still interested in merging this?

gnosis/eth/clients/blockscout_client.py
Kuzirashi: I'm not sure and I don't have time to test it unfortunately. I just updated old one which is probably deprecated to new one.

Kuzirashi:
> @Kuzirashi are you still interested in merging this?

I think it would be an improvement to the current codebase because new network would actually work. Now it doesn't.

gnosis/eth/clients/blockscout_client.py
Uxio0: Please make sure, if not remove it

Uxio0:
> I think it would be an improvement to the current codebase because new network would actually work. Now it doesn't.

Sure, then try to be responsive here so we can fix the issues and merge it asap

Uxio0:
Close due to inactivity

PR #589, opened 2023-07-13
feat: adding Oasis Sapphire and Oasis Sapphire Testnet support

ElvisKrop:
Please add Oasis Sapphire and Oasis Sapphire Testnet addresses:
- https://github.com/safe-global/safe-deployments/pull/221
- https://github.com/safe-global/safe-deployments/pull/238

Thank you!

### DIFF
--- a/gnosis/safe/addresses.py
+++ b/gnosis/safe/addresses.py
@@ -526,6 +526,14 @@ MASTER_COPIES: Dict[EthereumNetwork, List[Tuple[str, int, str]]] = {
         ("0xfb1bffC9d739B8D520DaF37dF666da4C687191EA", 1408599, "1.3.0+L2"),
         ("0x69f4D1788e39c87893C980c06EdF4b7f686e2938", 1408613, "1.3.0"),
     ],
+    EthereumNetwork.OASIS_SAPPHIRE: [
+        ("0xfb1bffC9d739B8D520DaF37dF666da4C687191EA", 325640, "1.3.0+L2"),
+        ("0x69f4D1788e39c87893C980c06EdF4b7f686e2938", 325643, "1.3.0"),
+    ],
+    EthereumNetwork.OASIS_SAPPHIRE_TESTNET: [
+        ("0xfb1bffC9d739B8D520DaF37dF666da4C687191EA", 1378154, "1.3.0+L2"),
+        ("0x69f4D1788e39c87893C980c06EdF4b7f686e2938", 1378155, "1.3.0"),
+    ],
 }
 
 PROXY_FACTORIES: Dict[EthereumNetwork, List[Tuple[str, int]]] = {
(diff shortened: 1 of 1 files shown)


moisses89:
@ElvisKrop I can check the contracts in the following blockexplorer https://explorer.sapphire.oasis.io/address/0xfb1bffC9d739B8D520DaF37dF666da4C687191EA/transactions

ElvisKrop:
> @ElvisKrop I can't check the contracts in the following blockexplorer https://explorer.sapphire.oasis.io/address/0xfb1bffC9d739B8D520DaF37dF666da4C687191EA/transactions

Hello @moisses89! Sorry for the delay, I was communicating with Oasis team to get some insights about this weird behaviour.
Unfortunately, Sapphire Explorer (Blockscout) can not know about contracts created by other contracts or used only as a library (e.g. only ever called from other contracts, but not via any transaction), because there are no transactions or events related to them. The only way to detect them is through EVM-tracing which would extract such information from the EVM state, but this has not been implemented yet.

However, you can explore [safe-singleton-factory transactions](https://explorer.sapphire.oasis.io/address/0x914d7Fec6aaC8cd542e72Bca78B30650d45643d7/transactions). There are exactly 10 transactions for contract creations:
1. SimulateTxAccessor
2. GnosisSafeProxyFactory
3. DefaultCallbackHandler
4. CompatibilityFallbackHandler
5. CreateCall
6. MultiSend
7. MultiSendCallOnly
8. SignMessageLib
9. GnosisSafeL2
10. GnosisSafe

_(note the reverse order of displayed transactions)_

image

Moreover, you can get the code from the RPC node directly using next request:
```
curl --location --request POST 'https://sapphire.oasis.io/' \
--header 'Content-Type: application/json' \
--data-raw '{
	"jsonrpc":"2.0",
	"method":"eth_getCode",
	"params":[
		"0xfb1bffC9d739B8D520DaF37dF666da4C687191EA",
                "latest"
	],
	"id":1
}'
```

The result is the same if you will perform the same request to the Polygon RPC node (https://polygon-rpc.com).
Thanks in advance!

gnosis/safe/addresses.py
```diff
@@ -522,6 +522,14 @@
         ("0xfb1bffC9d739B8D520DaF37dF666da4C687191EA", 1511, "1.3.0+L2"),
         ("0x69f4D1788e39c87893C980c06EdF4b7f686e2938", 1512, "1.3.0"),
     ],
+    EthereumNetwork.OASIS_SAPPHIRE: [
+        ("0xfb1bffC9d739B8D520DaF37dF666da4C687191EA", 330092, "1.3.0+L2"),
+        ("0x69f4D1788e39c87893C980c06EdF4b7f686e2938", 330094, "1.3.0"),
```
moisses89, line 527: @ElvisKrop thanks for your explanation here https://github.com/safe-global/safe-eth-py/pull/589#issuecomment-1641711020 My concern is not the code, because it was validated on safe-deployments PR, my concern is when was created, taking a look to your picture seems that the blocks that are harcoded here are incorrect, this block should be the creation block. I'd suggest 325640 and 325643.

moisses89 (changes requested):
Please review the blocks, should be the block where the contract was created.

gnosis/safe/addresses.py
ElvisKrop: omg, sorry for that...
I was focused on the wrong topic :sweat_smile: haha
thanks for pointing! I'll update them shortly

ElvisKrop: @moisses89 
:heavy_check_mark: blocks for oasis sapphire mainnet contracts are updated
:heavy_check_mark: I can confirm that block for testnet match the creation blocks of each contract too
:heavy_check_mark: PR branch is rebased

Hopefully, it is the final re-request for review :smile:
Thank you!

moisses89 (approved):
Looks good to me!

Issue #583, opened 2023-07-10
Add support for safe 1.4.1


moisses89:
# What is needed?
Add  support for safe  [1.4.1 ](https://github.com/safe-global/safe-contracts/releases/tag/v1.4.1)

# Taks
- [x]  #619 
- [x] #620 
- [x] #630

Uxio0:
Support for contracts v1.4.1 has been completed

PR #597, opened 2023-07-28
Refactor proxyFactory and safe

### LINKED ISSUE #619: Refactor safe and proxy factory
# What is needed?
I want to implement inheritance for the `Safe` class and the `ProxyFactory` class in order to maintain compatibility with older Safe contract versions, such as 1.3.0, while simultaneously introducing new Safe contract versions like 1.4.1.

# Related issues
#583

moisses89:
# Description 
The purpose of this pull request is refactor proxyFactory and safe class to do more easy add new safe contract versions. 

# Actions
- split proxy factory tests by version
- add a contract class with the common methods to deploy and get contract information 
- refactor proxy factory adding heritage with different proxy factory versions. 
- refactor safe class adding heritage with different safe versions.
# Safe refactor class 
Class diagram 
SafeClass drawio

The Safe class get the version from blockchain and decide which class should instantiate taking account the version and return the safe instance with the correct version. 
safe = Safe("[0x4E24f2497445125C4573f7935ADeDEce154317b8](https://gnosisscan.io/address/0x4e24f2497445125c4573f7935adedece154317b8#readProxyContract#F2)", ethereum_client) # returns instance of SafeV111
safe = Safe("[0xfca7Da0a0290D7BcBEcD93bE124756fC9B6F8E6A](https://gnosisscan.io/address/0xfca7da0a0290d7bcbecd93be124756fc9b6f8e6a#readProxyContract#F1)", ethereum_client) # returns instance of SafeV130

SafeV100 overwrite common method `retrieve_all_info` because the Safe V1.1.1 contract doesn't have getModulesPaginated. 

# Considerations
In case that Safe class couldn't get the version from blockchain, it would return the last version class to avoid break some tests.  We could improve this raising an exception that is an invalid safe. 
In case of version contains wrong version value we are returning SafeV130 to keep going with the unsupported mastercopies. 
 
 
# Issues related 
#583 
Closes #619


Uxio0 (changes requested):
PR is quite good, but I think we can still refactor more common things into the new classes

Also, afterwards we should reevaluate estimations and other methods to know if they are still valid of they need to be adjusted

gnosis/eth/contract_common.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @staticmethod
+    def deploy_contract_with_deploy_function(
+        ethereum_client: EthereumClient,
+        deployer_account: LocalAccount,
+        deploy_function: ContractFunction,
+        gas: Optional[int] = None,
+        gas_price: Optional[int] = None,
+        nonce: Optional[int] = None,
+        gas_increment: Optional[int] = None,
```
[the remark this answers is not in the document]
moisses89: I keep the increments because some tests were failing, but we can analyze why.

gnosis/safe/proxy_factory.py
```diff
@@ -191,34 +103,104 @@ def deploy_proxy_contract_with_nonce(
...
-        contract_address = create_proxy_fn.call(tx_parameters)
+        # Auto estimation of gas does not work. We use a little more gas just in case (gas_increment)
+        return self.deploy_contract_with_deploy_function(
+            self.ethereum_client,
+            deployer_account,
+            create_proxy_fn,
+            gas,
+            gas_price,
+            nonce,
+            gas_increment=50000,
```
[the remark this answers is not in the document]
moisses89: The tests are failing because the gas increment https://github.com/safe-global/safe-eth-py/actions/runs/6073638501/job/16475931902?pr=597#step:7:101

gnosis/safe/safe.py
```diff
@@ -1185,3 +813,453 @@ def send_multisig_tx(
         )
 
         return EthereumTxSent(tx_hash, tx, None)
+
+
+class SafeV130(SafeCommon):
+    @cached_property
+    def contract(self) -> Contract:
+        return get_safe_V1_3_0_contract(self.w3, address=self.address)
```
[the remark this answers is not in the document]
moisses89: If we add `contract_fn = get_safe_V0_0_1_contract` for class SafeV001 or the same for other classes is going to fail when we call `self.contract_fn` because `get_safe_V0_0_1_contract` is not part of safe class and we are passing the instance.

gnosis/safe/proxy_factory.py
```diff
@@ -191,34 +103,104 @@ def deploy_proxy_contract_with_nonce(
...
+        proxy_factory_contract = get_proxy_factory_V1_0_0_contract(ethereum_client.w3)
+        return cls.deploy_contract(
+            ethereum_client, deployer_account, proxy_factory_contract
         )
-        return EthereumTxSent(tx_hash, tx, contract_address)
 
     def get_contract(self, address: Optional[ChecksumAddress] = None):
         address = address or self.address
-        return get_proxy_factory_contract(self.ethereum_client.w3, address)
+        return get_proxy_factory_V1_0_0_contract(self.ethereum_client.w3, address)
```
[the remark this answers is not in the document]
moisses89: This is failing, because get_proxy_factory_... is not receiving the instance as argument. We can work around of it at the following way:
```
    @staticmethod
    def get_proxy_factory_fn(w3: Web3, address: Optional[str] = None):
        return get_proxy_factory_V1_3_0_contract(w3, address)
```

gnosis/eth/contract_common.py
```diff
@@ -0,0 +1,86 @@
...
+    ):
+
```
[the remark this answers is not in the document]
moisses89: I decided to revert the change, I think don't make sense keep that in contract_common

moisses89:
@Uxio0 created following PR to fix the test issues with blockscout https://github.com/safe-global/safe-eth-py/pull/621

gnosis/eth/ethereum_client.py
```diff
@@ -1698,6 +1704,57 @@ def get_blocks(
...
+        if gas_price is not None:
+            tx_params["gasPrice"] = gas_price
+
+        if gas is not None:
+            tx_params["gas"] = gas
+
+        if nonce is not None:
+            tx_params["nonce"] = nonce
+
+        if chain_id is not None:
```
moisses89, line 1753: if `chain_id` should be enough I think the same for above conditions.

gnosis/safe/safe.py
```diff
@@ -207,423 +139,130 @@ def _deploy_master_contract(
...
-    @staticmethod
-    def deploy_master_contract_v1_0_0(
-        ethereum_client: EthereumClient, deployer_account: LocalAccount
-    ) -> EthereumTxSent:
-        """
-        Deploy master contract. Takes deployer_account (if unlocked in the node) or the deployer private key
+        safe_tx_gas = estimated_tx_gas
+        base_gas = 0
+        gas_price = 1
+        gas_token = gas_token or NULL_ADDRESS
```
moisses89, line 227: Can be the default NULL_ADDRESS value defined in the method header?

gnosis/safe/safe.py
```diff
@@ -1061,127 +701,445 @@ def retrieve_threshold(
...
+        :param safe_tx_gas: Gas that should be used for the Safe transaction
+        :param base_gas: Gas costs for that are independent of the transaction execution
+            (e.g. base transaction fee, signature check, payment of the refund)
+        :param gas_price: Gas price that should be used for the payment calculation
+        :param gas_token: Token address (or `0x000..000` if ETH) that is used for the payment
+        :param refund_receiver: Address of receiver of gas payment (or `0x000..000` if tx.origin).
+        :param signatures: Packed signature data ({bytes32 r}{bytes32 s}{uint8 v})
+        :param safe_nonce: Nonce of the safe (to calculate hash)
+        :param safe_version: Safe version (to calculate hash)
+        :return:
```
moisses89, line 735: return: safe transaction

moisses89 (commented):
Looks good to me, some comments :)

gnosis/eth/contracts/__init__.py
```diff
@@ -74,6 +64,19 @@ def _load_json_file(path):
...
+
+def _abi_file_path(file):
+    return os.path.abspath(os.path.join(os.path.dirname(__file__), "abis", file))
+
+
+def _load_json_file(path):
+    with open(path) as f:
+        return json.load(f)
+
+
```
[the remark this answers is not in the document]
Uxio0: Done, better late than nothing

gnosis/eth/ethereum_client.py
Uxio0: Sorry, I'm not understanding

gnosis/safe/safe.py
Uxio0: This could break some behaviour, also `NULL_ADDRESS` is our current implementation for Solidity, while someone that's just using Python can say that the `gas_token` is `None`, and that should be fair in my opinion. We are just "translating"

Uxio0: We can change that but I wouldn't do it in this PR

gnosis/safe/safe.py
Uxio0: Done

Uxio0:
Thanks @moisses89 for starting this PR and all the contributions

PR #632, opened 2023-09-14
Add safe-deployments

### LINKED ISSUE #622: Add safe-deployment addresses
# What is needed?
Python users should have an easy way to access Safe deployments trusted addresses on https://github.com/safe-global/safe-deployments/tree/main/src/assets/ . We must implement:
- A script to update the addresses from **safe-deployments** repository (similar to the one we have for chain ids)
- A file for all the addresses for the related Safe contracts

# Background
Currently we manually maintain `safe_addresses.py` only with the addresses important to the safe-tx-service , and also some others besides the official ones. We should not deprecate that but should think about it.

# Related issues
- https://github.com/safe-global/safe-transaction-service/issues/1636

moisses89:
Closes #622 
- Add `safe_deployments.py `
- Add script to generate `safe_deployments.py` from https://github.com/safe-global/safe-deployments
- Move chains.py and load_safe_deployments.py to generators folder


scripts/load_safe_deployments.py
```diff
@@ -0,0 +1,75 @@
+import json
+
+import requests
+
+safe_deployments_base_url = (
+    "https://raw.githubusercontent.com/safe-global/safe-deployments/main/src/assets"
+)
+safe_deployments_path = "../gnosis/safe/safe-deployments/safe-deployments.json"
+
+deployments = [
```
Uxio0, line 10: Don't hardcode all this. Open the directory (maybe you need to clone it) and check the files inside, using something like `os.walk()`. That way we don't have to adapt the script if there's a new release or a new file is added

scripts/load_safe_deployments.py
```diff
@@ -0,0 +1,75 @@
+import json
+
+import requests
+
+safe_deployments_base_url = (
+    "https://raw.githubusercontent.com/safe-global/safe-deployments/main/src/assets"
+)
+safe_deployments_path = "../gnosis/safe/safe-deployments/safe-deployments.json"
```
Uxio0, line 8: Maybe rename to `output_file` or add a comment pointing out that this will be the output file

scripts/load_safe_deployments.py
```diff
@@ -0,0 +1,75 @@
...
+    for deployment in deployments:
+        for contract_filename in deployment["contracts_json"]:
+            response = requests.get(
+                f"{safe_deployments_base_url}/{deployment['version']}/{contract_filename}"
+            )
+            safe_deployment = response.json()
+            safe_deployments[safe_deployment["version"]] = {
+                safe_deployment["contractName"]: safe_deployment["networkAddresses"]
+            }
+    # Write json file
```
Uxio0, line 69: I would generate the file in `Python` instead of `json`, so we don't need to load it when using it

Uxio0 (changes requested):

scripts/load_safe_deployments.py
moisses89: Yes I was thinking the same, this PR is still in draft and working in progress, anyway I appreciate your suggestions :)

scripts/load_safe_deployments.py
```diff
@@ -0,0 +1,44 @@
+import json
+import os
+import shutil
+
+from git import Repo
+
+GIT_URL = "https://github.com/safe-global/safe-deployments.git"
+SAFE_DEPLOYMENTS_VERSION = "v1.27.0"  # safe-deployments tag version
+REPO_DIR = "safe-deployments"  # temporary folder to clone the repo
+DEPLOYMENTS_FOLDER = REPO_DIR + "/src/assets"  # folder where are the deployments
```
Uxio0, line 10:
```suggestion
DEPLOYMENTS_FOLDER = REPO_DIR + "/src/assets"  # folder where the deployments are
```

scripts/load_safe_deployments.py
```diff
@@ -0,0 +1,44 @@
...
+SAFE_ETH_PY_DEPLOYMENTS_PATH = "../gnosis/safe/safe_deployments.py"  # Path and filename where the new dict is being stored
```
Uxio0, line 11:
```suggestion
SAFE_ETH_PY_DEPLOYMENTS_PATH = "../gnosis/safe/safe_deployments.py"  # Full path where the deployment dictionary will be stored
```

scripts/load_safe_deployments.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def clean_resources():
```
Uxio0, line 14: Not required, but you could use a custom context manager to make sure resources are always cleaned: https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager

scripts/load_safe_deployments.py
```diff
@@ -0,0 +1,44 @@
...
+    """Removes the intermediate resources used (source repository)"""
+    try:
+        shutil.rmtree(REPO_DIR)
+    except OSError:
+        pass
+
+
+def generate_safe_deployments_py():
+    """
+    Get safe deployments addresses from https://github.com/safe-global/safe-deployments/ and generate python dict file
```
Uxio0, line 24:
```suggestion
    Get safe deployments addresses from https://github.com/safe-global/safe-deployments/ and generate python dictionary file
```

scripts/load_safe_deployments.py
```diff
@@ -0,0 +1,44 @@
+import json
```
Uxio0, line 1: Maybe we should move this script and chains one to a common and descriptive folder like `scripts/generators`? What do you think?

Uxio0 (changes requested):

scripts/load_safe_deployments.py
moisses89: Agree

Uxio0 (approved):

PR #652, opened 2023-10-09
Remove attrdict Web3 middleware

Uxio0:
- Not used by us
- Makes tests comparing dictionaries more difficult to understand
- Adds overhead when parsing information from the RPC


moisses89 (approved):
Looks good to me, I couldn't review every change in the mock files but I guess that is ok because it is passing tests.

PR #654, opened 2023-10-10
Support Safe Singleton Factory

Uxio0:
- Add deterministic contract deployment

Optimizes testing:
- Speed up tests, as it's easy to check if contracts are deployed
- Less confusing, as contract addresses will be the same to production environments

### DIFF
--- a/gnosis/eth/ethereum_client.py
+++ b/gnosis/eth/ethereum_client.py
@@ -57,6 +57,7 @@ from gnosis.eth.utils import (
     fast_is_checksum_address,
     fast_to_checksum_address,
     mk_contract_address,
+    mk_contract_address_2,
 )
 from gnosis.util import cache, chunks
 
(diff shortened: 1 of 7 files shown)


falvaradorodriguez (approved):

gnosis/eth/ethereum_client.py
```diff
@@ -1430,22 +1450,54 @@ def batch_call_same_function(
...
                 tx: TxParams = {
                     "from": deployer_account.address,
                     "data": data,
                     "gasPrice": self.w3.eth.gas_price,
                     "value": Wei(0),
                     "to": contract_address if contract_address else "",
                     "chainId": self.get_chain_id(),
+                    "nonce": self.get_nonce_for_account(deployer_account.address),
                 }
+                if not contract_address:
```
moisses89, line 1481: `contract_address` is always `None`

gnosis/eth/tests/ethereum_test_case.py
```diff
@@ -45,7 +45,8 @@ def setUpClass(cls):
         cls.multicall = cls.ethereum_client.multicall
         assert cls.multicall, "Multicall must be defined"
 
-    def get_ethereum_test_account(self) -> LocalAccount:
+    @classmethod
+    def get_ethereum_test_account(cls):
```
moisses89, line 49:
```suggestion
    def get_ethereum_test_account(cls) -> LocalAccount:
```
I'm not sure if should be a `classmethod` because is not using anything from class.

moisses89 (approved):

gnosis/eth/tests/ethereum_test_case.py
Uxio0: It's used from another `classmethod`, so it has to be a `classmethod`

PR #689, opened 2023-11-13
Calculate proxy expected address in ProxyFactory

Uxio0:
- Add `calculate_proxy_address` to ProxyFactory
- Add option to deploy specific chain id proxies (feature added in v1.4.1)

### DIFF
--- a/gnosis/safe/proxy_factory.py
+++ b/gnosis/safe/proxy_factory.py
@@ -2,6 +2,7 @@ import secrets
 from abc import ABCMeta
 from typing import Callable, Optional
 
+from eth_abi.packed import encode_packed
 from eth_account.signers.local import LocalAccount
 from eth_typing import ChecksumAddress
 from web3 import Web3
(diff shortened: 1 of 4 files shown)


gnosis/safe/proxy_factory.py
```diff
@@ -57,6 +66,36 @@ def deploy_contract(
...
+        """
+        :return: Creation code used for the Proxy deployment.
+            With this it is easily possible to calculate predicted address.
+        """
+        return self.contract.functions.proxyCreationCode().call()
+
+    @cache
+    def get_proxy_runtime_code(self) -> Optional[bytes]:
+        """
+        :return: Runtime code of a deployed Proxy. For v1.4.1 onwards the method is not available, so `None`
```
moisses89, line 80: For 1.4.1 the method was deprecated? If yes why not overwrite the function on 1.4.1 and raise deprecated exception?

gnosis/safe/proxy_factory.py
```diff
@@ -57,6 +66,36 @@ def deploy_contract(
...
+            will be returned
+        """
+        if hasattr(self.contract.functions, "proxyRuntimeCode"):
+            return self.contract.functions.proxyRuntimeCode().call()
+
+    def get_deploy_function(self, chain_specific: bool) -> ContractFunction:
+        if chain_specific:
```
moisses89, line 87: If this feature is available only in 1.4.1, why not implement it in the 1.4.1 class?

moisses89 (commented):
I think some of the methods could be written on specific class for version instead in a general way with conditions.

gnosis/safe/proxy_factory.py
Uxio0: Because this is required for the `check_proxy` method to work, we would need to refactor `check_proxy_code`. Also this was present on the previous version and it only shows as I sorted the methods to clean up the class

gnosis/safe/proxy_factory.py
Uxio0: Good point. I was trying to make something compatible across all the ProxyFactories, but you are right

moisses89 (approved):

PR #697, opened 2023-11-21
Fix: `run_tests.sh` script

### LINKED ISSUE #696: Error running tests: `ganache-cli` fails to run (?)
**Describe the bug**
I'm using the latest version of ganache-cli (Ganache CLI v6.12.2 (ganache-core: 2.13.2)), and the tests fail because they cannot connect to the node.

When I check the logs, I see `Unknown argument: chain` in the first line (I assume that's ganache's output).

Also, at the end of the test output:
```
Killing opened Ganache-Cli
./run_tests.sh: line 27: kill: (72561) - No such process
```

**To Reproduce**
1. clone the project
2. Install ganache globally
3. execute `./run_tests.sh`

**Expected behavior**
Tests pass

mmv08:
This PR:
- Do not rely on the globally installed `ganache-cli` package, use docker service
- Fixes #696

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,3 +7,12 @@ services:
       - "5432:5432"
     environment:
       POSTGRES_PASSWORD: postgres
+  ganache:
+    image: trufflesuite/ganache:latest
+    ports:
+      - "8545:8545"
+    command: --defaultBalanceEther 10000 --gasLimit 10000000 -a 30 --chain.chainId 1337 --chain.networkId 1337 -d
+    networks:
+      default:
+        aliases:
+          - ganache
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -1,14 +1,6 @@
 #!/bin/sh
 # Postgresql and ganache-cli must be running for the tests
 
-ps aux | grep ganache-cli | grep -v grep > /dev/null
-if [ $? -eq 1 ]; then
-    echo 'Running Ganache-Cli'
-    ganache-cli --defaultBalanceEther 10000 --gasLimit 10000000 -a 30 --chain.chainId 1337 --chain.networkId 1337 -d > /dev/null &
-    GANACHE_PID=$!
-    sleep 3
-fi
-
 docker ps | grep '\->5432' > /dev/null
 if [ $? -eq 1 ]; then
     docker-compose up -d db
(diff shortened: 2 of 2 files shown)


mmv08:

run_tests.sh
```diff
@@ -1,12 +1,12 @@
 #!/bin/sh
 # Postgresql and ganache-cli must be running for the tests
 
-ps aux | grep ganache-cli | grep -v grep > /dev/null
+ps aux | grep ganache | grep -v grep > /dev/null
 if [ $? -eq 1 ]; then
     echo 'Running Ganache-Cli'
-    ganache-cli --defaultBalanceEther 10000 --gasLimit 10000000 -a 30 --chain.chainId 1337 --chain.networkId 1337 -d > /dev/null &
+    npx ganache@7.9.1 --defaultBalanceEther 10000 --gasLimit 10000000 -a 30 --chain.chainId 1337 --chain.networkId 1337 -d > /dev/null &
```
Uxio0, line 7: Could you set latest instead of pinning an address? That's what we do in the CI and I don't want to go manually checking this script

mmv08: I think setting the latest is prone to errors on major releases. If some CLI flag is not recognized, it will not error and continue executing, and it's hard to catch. What do you think about putting the latest minor release?

The motivation for pinning a version is that I ran into errors because I was using an older version, and there needed to be info about what version I should use.

mmv08: For example, if you have ganache v6 installed, the tests will run but fail, and it took me a while to realise what was wrong,

Uxio0: > I think setting the latest is prone to errors on major releases. If some CLI flag is not recognized, it will not error and continue executing, and it's hard to catch. What do you think about putting the latest minor release?

I think then we will have to inform the user ganache is not executing and don't launch the tests. I want errors related to RPC to be detected as soon as we can, and using the latest ganache version for us is mandatory. I don't want to be checking this script from time to time. Also maybe we could remove `npx` dependency here and add `ganache` to `docker-compose`

mmv08: > lso maybe we could remove npx dependency here and add ganache to docker-compose

Technically npx is a transitive dependency of ganache because it has to be installed with npm, and npx is bundled with npm, so it's fine. I do like the idea of adding ganache to the docker image as it seems to be more aligned with how the CI works.

mmv08: Updated boss @Uxio0

docker-compose.yml
```diff
@@ -7,3 +7,12 @@ services:
       - "5432:5432"
     environment:
       POSTGRES_PASSWORD: postgres
+  ganache:
+    image: trufflesuite/ganache:latest
+    ports:
+      - "8545:8545"
+    command: --defaultBalanceEther 10000 --gasLimit 10000000 -a 30 --chain.chainId 1337 --chain.networkId 1337 -d
+    networks:
```
Uxio0, line 15: No need for a new network, please remove this

run_tests.sh
```diff
@@ -18,15 +10,19 @@ else
     DATABASE_UP=0
 fi
 
+docker ps | grep '\->8545' > /dev/null
```
Uxio0, line 13: We should start everything together, and then sleep for both

`docker-compose up -d ganache db`

Uxio0 (approved):

PR #707, opened 2023-12-05
Refactor Sourcify Client

Uxio0:
- Rename `Sourcify` to `SourcifyClient`
- Add method to check networks not supported
- Raise a `SourcifyClientConfigurationProblem` if network is not supported

### DIFF
--- a/gnosis/eth/clients/sourcify_client.py
+++ b/gnosis/eth/clients/sourcify_client.py
@@ -3,12 +3,22 @@ from urllib.parse import urljoin
 
 import requests
 
+from gnosis.util import cache
+
 from .. import EthereumNetwork
 from ..utils import fast_is_checksum_address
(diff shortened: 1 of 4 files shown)


gnosis/eth/clients/sourcify_client.py
```diff
@@ -63,6 +78,13 @@ def _do_request(self, url: str) -> Optional[Dict[str, Any]]:
 
         return response.json()
 
+    def is_chain_supported(self, chain_id: int) -> bool:
+        return chain_id in (int(chain["chainId"]) for chain in self.get_chains())
+
+    @cache
+    def get_chains(self) -> Dict[str, Any]:
+        return self._do_request("https://sourcify.dev/server/chains")
```
moisses89, line 86: This url could be better in a constant.

moisses89 (approved):

PR #721, opened 2023-12-19
Add support for preimage hash for EIP1271 signatures

Uxio0:


gnosis/safe/safe_signature.py
```diff
@@ -251,7 +274,7 @@ def is_valid(self, ethereum_client: EthereumClient, safe_address: str) -> bool:
             try:
                 return (
                     safe_contract.functions.approvedHashes(
-                        self.owner, self.safe_tx_hash
+                        self.owner, self.safe_hash_preimage
```
mmv08, line 277: This shouldn't change. A bytes32 hash is still used https://github.com/safe-global/safe-contracts/blob/main/contracts/Safe.sol#L342

Uxio0: I was testing and touched this, I was not meaning to, good catch!

PR #722, opened 2023-12-21
Add method to export signatures

Uxio0:
- This method is really relevant to build proper EIP1271 compatible signatures

### DIFF
--- a/gnosis/safe/safe_signature.py
+++ b/gnosis/safe/safe_signature.py
@@ -1,7 +1,7 @@
 from abc import ABC, abstractmethod
 from enum import Enum
 from logging import getLogger
-from typing import List, Optional, Union
+from typing import List, Optional, Sequence, Union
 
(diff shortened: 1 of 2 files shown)


gnosis/safe/safe_signature.py
```diff
@@ -148,10 +148,42 @@ def parse_signature(
             safe_signatures.append(safe_signature)
         return safe_signatures
 
+    @classmethod
+    def export_signatures(cls, safe_signatures: Sequence["SafeSignature"]) -> HexBytes:
```
moisses89, line 152: Why this name? In my opinion is too generic, could be better build_eip1271_signature?

Uxio0: That's not building a `1271` signature. It's building a valid and accepted signature by the smart contract

moisses89 (approved):

PR #749, opened 2024-01-24
Remove goerli from tests

### LINKED ISSUE #734: Migrate safe transaction API tests
# What is needed?
Our current setup tests for Safe transaction API requests were done with [Goerli](https://github.com/safe-global/safe-eth-py/blob/master/gnosis/safe/tests/api/test_transaction_service_api.py#L14). 
Goerli is going to be deprecated and Goerli Safe transaction service will be deprecated also, that means, this tests will fail.

# Proposed solution
We can migrate this tests to Sepolia, but I would prefer a more long life network, for example Gnosis Chain.

Uxio0:
- Remove goerli from e2e tests, use Gnosis Chain
- Add Sepolia API for cowswap
- Closes #734


moisses89 (approved):

Uxio0:
@moisses89 just updated the PR changing e2e tests from goerli to Gnosis Chain as suggested

PR #752, opened 2024-01-26
Fix api test

moisses89:
The Safe that we are testing have 7 transactions.


Uxio0:
This will not be required due to https://github.com/safe-global/safe-transaction-service/pull/1842

PR #761, opened 2024-01-31
feat: add Zora Mainnet 1.3.0 safe contracts deployed via safe singleton factory

ElvisKrop:
Thanks in advance!


falvaradorodriguez:
@ElvisKrop, the PR changes are correct. However, the BlockscoutClient URL (https://github.com/safe-global/safe-eth-py/blob/c2be195cece058a2d42db065e36a27c6ece35d02/gnosis/eth/clients/blockscout_client.py#L84) is incorrect from a previous commit. Please could you update it? I think the correct url should be: https://explorer.zora.energy/graphiql

falvaradorodriguez (approved):

PR #800, opened 2024-02-26
Refactor UserOperation constructor

Uxio0:
- Cache BundlerAPI

### DIFF
--- a/gnosis/eth/account_abstraction/user_operation.py
+++ b/gnosis/eth/account_abstraction/user_operation.py
@@ -1,13 +1,18 @@
 import dataclasses
-from functools import cached_property
 from typing import Any, Dict, Optional, Union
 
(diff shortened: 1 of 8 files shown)


gnosis/eth/account_abstraction/user_operation.py
```diff
@@ -18,6 +23,7 @@ class UserOperation:
     https://github.com/eth-infinitism/account-abstraction/blob/v0.6.0
     """
 
+    user_operation_hash: bytes
```
moisses89, line 26: We should use both bytes or str or decide what we want to use https://github.com/safe-global/safe-eth-py/pull/800/files#diff-ba5b3a9d52831b6f1bfb7059bfae7f9a0b90835e3b6d99855adcbafa88261171R84

moisses89 (approved):

PR #802, opened 2024-02-27
Rename enum SafeOperation to SafeOperationEnum

Uxio0:
- Remove confusion between 4337 SafeOperation and the enum for SafeTx


moisses89 (approved):
Makes sense

PR #818, opened 2024-03-05
Calculate SafeOperation preimage

Uxio0:
- It's needed for validating EIP1271 signatures

### DIFF
--- a/gnosis/safe/account_abstraction/safe_operation.py
+++ b/gnosis/safe/account_abstraction/safe_operation.py
@@ -54,6 +54,7 @@ class SafeOperation:
         "0x47e79534a245952e8b16893a336b85a3d9ea9fa8c573f3d803afb92a79469218"
     )  # bytes32
     safe_operation_hash: Optional[bytes] = None
+    safe_operation_hash_preimage: Optional[bytes] = None
 
     @classmethod
     def from_user_operation(cls, user_operation: UserOperation):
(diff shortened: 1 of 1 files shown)


gnosis/safe/account_abstraction/safe_operation.py
```diff
@@ -87,11 +88,11 @@ def get_domain_separator(
             )
         return _domain_separator_cache[key]
 
-    def get_safe_operation_hash(
+    def get_safe_operation_hash_preimage(
```
moisses89, line 91: Could be cached_property and avoid to check if exists the pre-image?

Uxio0: Dataclasses are not hashable by default, so cache decorators cannot be used. Also a property cannot be used as it receives parameters

Uxio0: Maybe we should use namedtuple for this case https://stackoverflow.com/questions/51671699/data-classes-vs-typing-namedtuple-primary-use-cases

moisses89: > Dataclasses are not hashable by default, so cache decorators cannot be used. Also a property cannot be used as it receives parameters

Got it.
My concern is if you create an instance of `UserOperation`, maybe you are not aware that the `useroperation_hash` is `None` until call the get function. 
Maybe we can add `chain` and `module` as atributes of `UserOperation` like  `DOMAIN_HASH_SEPARATOR` is and remove `safe_operation_hash` and `safe_operation_hash_preimage` as parameters and create the property functions.

Uxio0: Chain and module are not available when retrieving the UserOperation, that's why I don't want to force to define then as properties as they are not mandatory for working with the UserOperation and only needed when calculating the hash

moisses89 (approved):

PR #832, opened 2024-03-13
Optimize keccak operations

Uxio0:
- Use `fast_keccak`, `fast_keccak_text`, `fast_is_checksum_address` when posible
- Use caching for keccak operations. Memory usage is better than CPU usage

### DIFF
--- a/gnosis/eth/tests/eip712/test_eip712.py
+++ b/gnosis/eth/tests/eip712/test_eip712.py
@@ -86,7 +86,7 @@ class TestEIP712(TestCase):
         payload["types"] = self.types
         self.assertEqual(
             eip712_encode_hash(payload).hex(),
-            "7c02fe79823722257b42ea95720e7dd31d51c3f6769dc0f56a271800dd030ef1",
+            "0x7c02fe79823722257b42ea95720e7dd31d51c3f6769dc0f56a271800dd030ef1",
         )
 
     def test_eip712_encode_hash_string_uint(self):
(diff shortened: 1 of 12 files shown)


falvaradorodriguez (approved):

gnosis/eth/tests/eip712/test_eip712.py
```diff
@@ -86,7 +86,7 @@ def test_eip712_encode_hash(self):
         payload["types"] = self.types
         self.assertEqual(
             eip712_encode_hash(payload).hex(),
-            "7c02fe79823722257b42ea95720e7dd31d51c3f6769dc0f56a271800dd030ef1",
+            "0x7c02fe79823722257b42ea95720e7dd31d51c3f6769dc0f56a271800dd030ef1",
```
moisses89, line 89: This is related to return an HexBytes. I think that this issue should be tagged as breaking change.

moisses89 (changes requested):

gnosis/eth/utils.py
```diff
@@ -20,13 +21,25 @@ def get_empty_tx_params() -> TxParams:
...
     """
     Calculates ethereum keccak256 using fast library `pysha3`
+
+    :param value:
+    :return: Keccak256 used by ethereum as `HexBytes`
+    """
+    return HexBytes(keccak_256(value).digest())
+
+
+def fast_keccak_text(value: str) -> Hash32:
```
moisses89, line 35:
```suggestion
def fast_keccak_text(value: str) -> HexBytes:
```
It is returning an object not string.

Uxio0: `Hash32` is a `bytes`, and `HexBytes` is a sublass of `bytes`

gnosis/eth/tests/eip712/test_eip712.py
Uxio0: Totally agree

moisses89 (approved):

PR #875, opened 2024-04-02
Add missing workflow steps validations

falvaradorodriguez:


.github/workflows/validate_new_address_issue_input_data.yml
```diff
@@ -61,7 +61,7 @@ jobs:
 
     - name: Install dependencies
       run: |
-        pip install safe-eth-py==6.0.0b18
+        pip install safe-eth-py
         pip install validators==0.22.0
```
moisses89, line 65: It is necessary this specific version?

falvaradorodriguez: I have just tested it and it seems that it is not necessary.

moisses89 (approved):

PR #888, opened 2024-04-08
feat: add default singleton on Base

Majorfi:
The default singleton address on Base was missing: 

GnosisSafeL2
https://basescan.org/address/0x3E5c63644E683549055b9Be8653de26E0B4CD36E#code

GnosisSafeProxyFactory:
https://basescan.org/address/0xa6b71e26c5e0845f74c812102ca7114b6a896ab2

With this, could you also indicate if any other steps/prs are required to remove the `This Safe Account was created with an unsupported base contract. The web interface might not work correctly. We recommend using the command line interface instead.` message ? Thank you!


moisses89:
> With this, could you also indicate if any other steps/prs are required to remove the This Safe Account was created with an unsupported base contract. The web interface might not work correctly. We recommend using the command line interface instead. message ? Thank you!

Thanks! to remove the message will be necessary to create a release of `safe-eth-py ` and deploy a new version of transaction service. We will handle it asap.

moisses89 (approved):

PR #894, opened 2024-04-11
4337: Add method to get deployed account from receipt

Uxio0:
- Use `AccountDeployed` event to check if the account was indeed deployed

### DIFF
--- a/gnosis/eth/account_abstraction/user_operation_receipt.py
+++ b/gnosis/eth/account_abstraction/user_operation_receipt.py
@@ -6,6 +6,7 @@ from hexbytes import HexBytes
 from web3.types import LogReceipt
 
 from gnosis.eth.account_abstraction.constants import (
+    ACCOUNT_DEPLOYED_TOPIC,
     DEPOSIT_EVENT_TOPIC,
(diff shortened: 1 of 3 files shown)


gnosis/eth/account_abstraction/user_operation_receipt.py
```diff
@@ -44,6 +45,17 @@ def from_bundler_response(
...
+    def get_deployed_account(self) -> Optional[ChecksumAddress]:
+        """
+        :return: Deployed account in case a new account was deployed
+        """
+        for log in self.logs:
+            if (
+                len(log["topics"]) == 3
+                and HexBytes(log["topics"][0]) == ACCOUNT_DEPLOYED_TOPIC
+            ):
+                return fast_to_checksum_address(log["topics"][2][-40:])
```
moisses89, line 57: Maybe we could add some link documentation why `[2][-40:]`

moisses89 (approved):

gnosis/eth/account_abstraction/user_operation_receipt.py
Uxio0: Done, thanks!

PR #900, opened 2024-04-12
Cache bundler chain_id method

Uxio0:


falvaradorodriguez (approved):

gnosis/eth/account_abstraction/bundler_client.py
```diff
@@ -108,6 +108,7 @@ def _get_user_operation_receipt_payload(
             "id": request_id,
         }
 
+    @cache
```
moisses89, line 111: We have other get_chain_id in EthereumClient. Why we are not using it here?
https://github.com/safe-global/safe-eth-py/blob/72b7ed6daded3aa98cd536eb157551365f903a09/gnosis/eth/ethereum_client.py#L1308

moisses89 (changes requested):

moisses89 (approved):

PR #901, opened 2024-04-12
Bump django version

Uxio0:


falvaradorodriguez (approved):

moisses89:
Currently our services are using django 5. Should we bump to django 5?

moisses89 (approved):

Uxio0:
> Currently our services are using django 5. Should we bump to django 5?

It will break compatibility with older Python versions

PR #902, opened 2024-04-12
add shyft mainnet transaction service URL

svanegmond:
It seems this is hardcoded.

Some of our users are asking for safe-cli access.


moisses89:
This PR is breaking the transaction service API for `safe.global`, as you can see we are just supporting for now  `safe.global` transaction services.
Closing this issue because this change is not planned.

Issue #784, opened 2024-02-19
Allow configuration of internal parameters


Uxio0:
For example, currently in multiple places we have HTTP connection pool configurations that could be refactored to get the configuration from environment variables

For example, current code from `EthereumClient`:

```python
class EthereumClient:
    def __init__(
        self,
        ethereum_node_url: URI = URI("http://localhost:8545"),
        provider_timeout: int = 15,
        slow_provider_timeout: int = 60,
        retry_count: int = 1,
        use_caching_middleware: bool = True,
        batch_request_max_size: int = 500,
    ):
```

to

```python
class EthereumClient:
    def __init__(
        self,
        ethereum_node_url: URI = URI("http://localhost:8545"),
        provider_timeout: int = int(os.environ.get("ETHEREUM_NODE_TIMEOUT", 15)),
        slow_provider_timeout: int = int(os.environ.get("ETHEREUM_NODE_SLOW_TIMEOUT", 60)),
        retry_count: int = int(os.environ.get("ETHEREUM_NODE_RETRY_COUNT", 1)),
        use_caching_middleware: bool = True,
        batch_request_max_size: int = int(os.environ.get("ETHEREUM_NODE_BATCH_REQUEST_SIZE", 500)),
    ):
```

or define a configuration class like

```python
class EthereumClientConfiguration:
```

To refactor the environment reading and configuration

PR #947, opened 2024-04-30
Allow configuration of internal parameters

falvaradorodriguez:
Closes https://github.com/safe-global/safe-eth-py/issues/784

**To be considered:**

-  Some of this task was done in these previous PRs:
     https://github.com/safe-global/safe-eth-py/pull/209
     https://github.com/safe-global/safe-eth-py/pull/641

- I have checked some timeouts that were hardcoded, I would like to check with the team what other settings are missing.


gnosis/safe/api/base_api.py
```diff
@@ -24,7 +25,7 @@ def __init__(
         network: EthereumNetwork,
         ethereum_client: Optional[EthereumClient] = None,
         base_url: Optional[str] = None,
-        request_timeout: int = 10,
+        request_timeout: int = int(os.environ.get("SAFE_BASE_API_REQUEST_TIMEOUT", 10)),
```
moisses89, line 28: I think that would be better define this timeout in the childs, intead here. 
I mean something like `SAFE_TRANSACTION_SERVICE_REQUEST_TIMEOUT`, and the same for the relay, but not sure if will be continued.

Uxio0: Relay is discontinued, we should remove it. But I agree on defining this variable for every child

falvaradorodriguez: Added in TransactionServiceApi. RelayServiceApi will be removed in this [PR](https://github.com/safe-global/safe-eth-py/pull/949).

Uxio0 (approved):

Uxio0:
@falvaradorodriguez Could you add documentation for this when it's merged?

falvaradorodriguez:
> @falvaradorodriguez Could you add documentation for this when it's merged?

@Uxio0 updated [here](https://github.com/safe-global/safe-eth-py/pull/947/commits/8c06d42686394b8b201df34038594553b67dbf5e)

PR #968, opened 2024-05-07
Set version 6.0.0b28

Uxio0:


falvaradorodriguez (approved):

setup.cfg
```diff
@@ -1,6 +1,6 @@
 [metadata]
 name = safe-eth-py
-version = 6.0.0b25
+version = 6.0.0b26
```
falvaradorodriguez, line 3: The version should be 6.0.0b28.

Issue #890, opened 2024-04-10
Refactor `get_transactions` method of `transaction_service_api` to allow sending query params


falvaradorodriguez:
Currently, it is not possible to send filtering parameters to the `get_transactions` method of the `transaction_service_api` client.

It would be necessary to add the parameters allowed by the API and thus have the same functionality in safe-eth-py.

PR #978, opened 2024-05-13
Allow query params to get_transactions api method

falvaradorodriguez:
Closes https://github.com/safe-global/safe-eth-py/issues/890

### DIFF
--- a/gnosis/safe/api/transaction_service_api/transaction_service_api.py
+++ b/gnosis/safe/api/transaction_service_api/transaction_service_api.py
@@ -1,6 +1,7 @@
 import logging
 import os
 from typing import Any, Dict, List, Optional, Tuple, Union
+from urllib.parse import urlencode
 
 from eth_account.signers.local import LocalAccount
 from eth_typing import ChecksumAddress, HexStr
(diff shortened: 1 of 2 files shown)


gnosis/safe/api/transaction_service_api/transaction_service_api.py
```diff
@@ -168,15 +169,23 @@ def get_safe_transaction(
...
         :return: a list of transactions for provided Safe
         """
-        response = self._get_request(
-            f"/api/v1/safes/{safe_address}/multisig-transactions/"
-        )
+        url = f"/api/v1/safes/{safe_address}/multisig-transactions/"
+
+        if kwargs:
+            query_string = urlencode(
+                {key: f"{value}" for key, value in kwargs.items() if value is not None}
```
Uxio0, line 184: No strong preference but `str(value)` might look more clear? I don't know

gnosis/safe/api/transaction_service_api/transaction_service_api.py
```diff
@@ -168,15 +169,23 @@ def get_safe_transaction(
             safe_tx.tx_hash = tx_hash
         return safe_tx, tx_hash
 
-    def get_transactions(self, safe_address: ChecksumAddress) -> List[Dict[str, Any]]:
+    def get_transactions(
+        self, safe_address: ChecksumAddress, **kwargs: Union[str, int, bool]
```
Uxio0, line 173:
```suggestion
        self, safe_address: ChecksumAddress, **kwargs: Dict[str, Union[str, int, bool]]
```

Complex to annotate that 😅 : https://rednafi.com/python/annotate_args_and_kwargs/

Uxio0 (approved):

gnosis/safe/api/transaction_service_api/transaction_service_api.py
falvaradorodriguez: My mistake, I didn't pay enough attention. Good catch :)

Issue #750, opened 2024-01-24
Remove delegate API client shouldn't sign the message


moisses89:
# What is needed?
The function [remove_delegate](https://github.com/safe-global/safe-eth-py/blob/master/gnosis/safe/api/transaction_service_api.py#L240) of transaction service api is signing the message with wallet provided by parameter, but this signature should happens outside of the `remove_delegate` function to let other types of wallets as hardware wallets sign the message.
I'd suggest that `remove_delegate` should just receive the message_hash and the signature.

Uxio0:
Makes total sense @moisses89 , good call

PR #981, opened 2024-05-16
Add method to remove delegate with provided signature

falvaradorodriguez:
Closes https://github.com/safe-global/safe-eth-py/issues/750


Uxio0:
@moisses89 when this is ready could you review it?

moisses89 (dismissed):
Looks good to me!

moisses89:
Just one question, should we deprecate this function?
https://github.com/safe-global/safe-eth-py/pull/981/files#diff-46696cb4d085f46df32fdb8c28828f6b36710fad85d3b6e36240c3c56c14005bL287

gnosis/safe/api/transaction_service_api/transaction_service_api.py
```diff
@@ -305,6 +305,37 @@ def remove_delegate(
             raise SafeAPIException(f"Cannot remove delegate: {response.content}")
         return True
 
+    def remove_delegate_signed(
```
moisses89, line 308: I don't like to have two functions to remove delegates. We have here two options in my opinion. 
- Remove the function that is signing inside
- or remove_delegate with optional signature and optional signature parameters as the signer. 
I prefer the first one, because in my opinion calculate the signature is out of the goal of a client API function.
For both options we should set the label breaking_change for the issue.

moisses89 (changes requested):

gnosis/safe/api/transaction_service_api/transaction_service_api.py
falvaradorodriguez: Ok, makes sense. Removed here https://github.com/safe-global/safe-eth-py/pull/981/commits/a7eb6923bf34637721a047a1946c4062057f4a15

falvaradorodriguez:
@moisses89 Seeing this it made sense to me to apply the same criteria for the add_delegate method. I have also put an optional safe parameter as in the case of delete. If you see that this doesn't apply here, please let me know and I'll revert it. Thanks

moisses89 (approved):

Uxio0 (approved):

Issue #972, opened 2024-05-09
subclassing safe doesn't work


banteg:
https://github.com/safe-global/safe-eth-py/blob/576d510ba87da1cb414857b622c56455bdf50897/gnosis/safe/safe.py#L95

```python
…/gnosis/safe/safe.py", line 96, in __new__
    return super().__new__(cls, address, ethereum_client, *args, **kwargs)
TypeError: object.__new__() takes exactly one argument (the type to instantiate)
```

Uxio0:
Could you provide some info? We subclass `Safe` without further issues, like: https://github.com/safe-global/safe-eth-py/blob/576d510ba87da1cb414857b622c56455bdf50897/gnosis/safe/safe.py#L981

banteg:
```python
In [1]: from gnosis.eth import EthereumClient

In [2]: from gnosis.safe.safe import SafeV130, Safe

In [3]: client = EthereumClient()

In [4]: addr = '0xFEB4acf3df3cDEA7399794D0869ef76A6EfAff52'

In [5]: class MySafe(Safe):
   ...:     pass
   ...:

In [6]: class MySafeV130(SafeV130):
   ...:     pass
   ...:

In [7]: Safe(addr, client)
Out[7]: <gnosis.safe.safe.SafeV130 at 0x107a9b790>

In [8]: SafeV130(addr, client)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[8], line 1
----> 1 SafeV130(addr, client)

File ~/yearn/brownie-safe/.venv/lib/python3.10/site-packages/gnosis/safe/safe.py:95, in Safe.__new__(cls, address, ethereum_client, *args, **kwargs)
     93 assert fast_is_checksum_address(address), "%s is not a valid address" % address
     94 if cls is not Safe:
---> 95     return super().__new__(cls, address, ethereum_client, *args, **kwargs)
     97 versions: Dict[str, Safe] = {
     98     "0.0.1": SafeV001,
     99     "1.0.0": SafeV100,
   (...)
    103     "1.4.1": SafeV141,
    104 }
    105 default_version = SafeV141

TypeError: object.__new__() takes exactly one argument (the type to instantiate)

In [9]: MySafe(addr, client)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[9], line 1
----> 1 MySafe(addr, client)

[...]

PR #992, opened 2024-05-29
Fix inheritance safe subclasses

falvaradorodriguez:
Closes https://github.com/safe-global/safe-eth-py/issues/972

There was a problem when passing parameters to the __new__ method of the superclass in the constructor of the Safe class. These parameters must be passed in the __init__ method but not in the __new__ method.

If you directly instantiated a class that inherits from Safe, such as SafeV141, instead of passing it through the __new__ constructor of the Safe class, the same problem occurred.

A test case is added to validate the explicit condition in the Safe constructor.


Uxio0 (approved):
Really nice job and Python understanding to implement this fix

Issue #1006, opened 2024-05-30
Add method to decode transaction data in Tx Service API


falvaradorodriguez:
Just as you can decode from the SDK's SafeAPIKit, it should be possible to decode from safe-eth-py.

PS: Update the documentation once this issue has been merged.

PR #1007, opened 2024-05-30
Add decode data method in Tx Service Api

falvaradorodriguez:
Closes https://github.com/safe-global/safe-eth-py/issues/1006


gnosis/safe/api/transaction_service_api/transaction_service_api.py
```diff
@@ -416,3 +416,21 @@ def post_message_signature(
...
         return True
+
+    def decode_data(
+        self, data: Union[bytes, HexStr], to_address: Optional[ChecksumAddress] = None
+    ) -> Dict[str, Any]:
+        """
+        Retrieve decoded information using tx service internal ABI information given the tx data.
+
+        :param data: tx data as a 0x prefixed hexadecimal string.
+        :param to_address: address of the receiving contract
```
Uxio0, line 427:
```suggestion
        :param to_address: address of the contract. This will be used in case of more than one function identifiers matching.
```

gnosis/safe/api/transaction_service_api/transaction_service_api.py
```diff
@@ -416,3 +416,21 @@ def post_message_signature(
...
+        :return:
```
Uxio0, line 428: What are you returning? Maybe you can define a TypedDict (inside a file `entities.py` on the `transaction_service_api/` folder) to make clear what's the return value. And we should do the same for other functions

gnosis/safe/tests/api/test_transaction_service_api.py
```diff
@@ -183,3 +184,15 @@ def test_get_safe_transaction(self):
...
                 str(context.exception),
             )
+
+    def test_decode_data(self):
+        with patch.object(TransactionServiceApi, "_post_request") as mock_post_request:
+            data = HexStr(
+                "0x095ea7b3000000000000000000000000e6fc577e87f7c977c4393300417dcc592d90acf8ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
+            )
+            to_address = "0x4127839cdf4F73d9fC9a2C2861d8d1799e9DF40C"
+            self.transaction_service_api.decode_data(data, to_address)
```
Uxio0, line 194: You are not checking the function is returning anything. I could change `return response.json()` in `decode_data` function to `return None` at the end and this test will pass. You should also mock the result and check the function is returning what you expect.

Uxio0 (changes requested):

gnosis/safe/api/transaction_service_api/transaction_service_api.py
falvaradorodriguez: You are right, I did this by looking at the other methods, but it is not the right thing to do. I will try to update the other method signatures.

Uxio0 (approved):
Nice job

PR #1015, opened 2024-05-31
Update safe-deployments to last version

Uxio0:
- Update script to make directories relative, so script can be called from any folder
- Rename constant names


scripts/generators/generate_safe_deployments.py
```diff
@@ -5,25 +5,30 @@
 
 from git import Repo
 
-GIT_URL = "https://github.com/safe-global/safe-deployments.git"
-SAFE_DEPLOYMENTS_VERSION = "v1.35.0"  # safe-deployments tag version
-REPO_DIR = "safe-deployments"  # temporary folder to clone the repo
-DEPLOYMENTS_FOLDER = REPO_DIR + "/src/assets"  # folder where the deployments are
-SAFE_ETH_PY_DEPLOYMENTS_PATH = "../../gnosis/safe/safe_deployments.py"  # Full path where the deployment dictionary will be stored
+SAFE_DEPLOYMENTS_URL = "https://github.com/safe-global/safe-deployments.git"
+SAFE_DEPLOYMENTS_VERSION = "main"  # safe-deployments tag/branch
```
falvaradorodriguez, line 9: Just to confirm: No need to release the changes, is it enough if they are in main?

falvaradorodriguez (approved):

scripts/generators/generate_safe_deployments.py
Uxio0: Yes, it should be alright

Issue #713, opened 2023-12-12
Add function to insert ordered signature to SafeTx 


moisses89:
# What is needed?
Currently some projects that uses the `safe-eth-py` library as `safe-cli` are keeping by themself the sorting of `safeTx` signatures each time that new signature is added to  `safeTx`, for example: 
```
 # Insert signature sorted
            if wallet.address not in safe_tx.signers:
                new_owners = safe_tx.signers + [wallet.address]
                new_owner_pos = sorted(new_owners, key=lambda x: int(x, 16)).index(
                    wallet.address
                )
                safe_tx.signatures = (
                    safe_tx.signatures[: 65 * new_owner_pos]
                    + signature
                    + safe_tx.signatures[65 * new_owner_pos :]
                )
```
I'd suggest include a function in `safeTx` class that lets a developer to just provide a signature or list of signatures and the function insert the new signature sorted in the `safeTx`.

PR #1060, opened 2024-06-05
Insert ordered signatures in Safe Tx

falvaradorodriguez:
Closes https://github.com/safe-global/safe-eth-py/issues/713

The approach of the task changes due a utility for sorting and exporting signatures has been created in this task https://github.com/safe-global/safe-eth-py/pull/722.

This implies that the responsibility for sorting belongs to this new utility and that it is not necessary to do the task as foreseen.

In the PR only the management that was duplicated in the SafeTx class is removed and the existing SafeSignature is used.

### DIFF
--- a/gnosis/safe/safe_tx.py
+++ b/gnosis/safe/safe_tx.py
@@ -415,22 +415,19 @@ class SafeTx:
 
         # Insert signature sorted
         if account.address not in self.signers:
-            new_owners = self.signers + [account.address]
-            new_owner_pos = sorted(new_owners, key=lambda x: int(x, 16)).index(
-                account.address
+            new_tx_signatures = SafeSignature.parse_signature(
+                self.signatures + signature, self.safe_tx_hash
             )
-            self.signatures = (
-                self.signatures[: 65 * new_owner_pos]
-                + signature
-                + self.signatures[65 * new_owner_pos :]
-            )
+            self.signatures = SafeSignature.export_signatures(new_tx_signatures)
+
         return signature
 
     def unsign(self, address: str) -> bool:
-        for pos, signer in enumerate(self.signers):
-            if signer == address:
-                self.signatures = self.signatures.replace(
-                    self.signatures[pos * 65 : pos * 65 + 65], b""
-                )
-                return True
+        tx_signatures = SafeSignature.parse_signature(
+            self.signatures, self.safe_tx_hash
+        )
+        new_tx_signatures = list(filter(lambda x: x.owner != address, tx_signatures))
+        if tx_signatures != new_tx_signatures:
+            self.signatures = SafeSignature.export_signatures(new_tx_signatures)
+            return True
         return False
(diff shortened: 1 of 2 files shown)


gnosis/safe/safe_tx.py
```diff
@@ -415,22 +415,19 @@ def sign(self, private_key: str) -> bytes:
...
         if account.address not in self.signers:
-            new_owners = self.signers + [account.address]
-            new_owner_pos = sorted(new_owners, key=lambda x: int(x, 16)).index(
-                account.address
-            )
-            self.signatures = (
-                self.signatures[: 65 * new_owner_pos]
-                + signature
-                + self.signatures[65 * new_owner_pos :]
+            new_tx_signatures = SafeSignature.parse_signature(
```
moisses89, line 418: Maybe just `signatures`, because `new` is not accurated to the variable, I mean the variable contains old signatures and new signatures. 
Other option could be unsorted_signatures is you want to remark that are unsorted and needs to be sorted.

gnosis/safe/safe_tx.py
```diff
@@ -415,22 +415,19 @@ def sign(self, private_key: str) -> bytes:
...
-        for pos, signer in enumerate(self.signers):
-            if signer == address:
-                self.signatures = self.signatures.replace(
-                    self.signatures[pos * 65 : pos * 65 + 65], b""
-                )
-                return True
+        tx_signatures = SafeSignature.parse_signature(
+            self.signatures, self.safe_tx_hash
+        )
+        new_tx_signatures = list(filter(lambda x: x.owner != address, tx_signatures))
```
moisses89, line 429: I would expect one signature by owner, why returns a list? 
Anyway, do you think that make sense change the `unsign`? in this case remove one signature from the middle does not affect the sort.

falvaradorodriguez: I think the problem is in the name of the method. The `signature` stores the signatures of all owners in bytes... that's why you need to parse them to get the list to manage it.

I did it to keep the logic at the same point. In this case `SafeSignature`. If you see that it doesn't help, I can remove this change, what do you say?

gnosis/safe/safe_tx.py
falvaradorodriguez: Updated here https://github.com/safe-global/safe-eth-py/pull/1060/commits/5c8cf1e79ba47af31cbe56a4ba54c60c69eca9a4

gnosis/safe/safe_tx.py
falvaradorodriguez: If I keep it, I must change the name of the variable to be consistent with the above.

falvaradorodriguez: > If I keep it, I must change the name of the variable to be consistent with the above.

Done here https://github.com/safe-global/safe-eth-py/pull/1060/commits/314d30b6b43cd30c8880d9573fd8b6819888b5a2

moisses89 (approved):

PR #1171, opened 2024-06-14
Dont respect retry_after_header

Uxio0:
- Previously, if a `RetryAfter` header is set it will overwrite our retry policy
- We should be the ones deciding on the retry policy
- More info on https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After

### DIFF
--- a/gnosis/util/http.py
+++ b/gnosis/util/http.py
@@ -15,8 +15,7 @@ def prepare_http_session(
     session = requests.Session()
     retry_conf = (
         requests.adapters.Retry(
-            total=retry_count,
-            backoff_factor=0.3,
+            total=retry_count, backoff_factor=0.3, respect_retry_after_header=False
         )
         if retry_count
         else 0


gnosis/util/http.py
```diff
@@ -15,8 +15,7 @@ def prepare_http_session(
     session = requests.Session()
     retry_conf = (
         requests.adapters.Retry(
-            total=retry_count,
-            backoff_factor=0.3,
+            total=retry_count, backoff_factor=0.3, respect_retry_after_header=False
```
fmrsabino, line 18: Out of curiosity, wouldn't the server know better when we should attempt to retry the connection?

What are the use cases that you are thinking of where this decision should be on the client side?

Uxio0: Maybe the service knows better in some scenarios. But for example some services tell you to retry in one hour, for example, blocking the request forever

fmrsabino (approved):

PR #1177, opened 2024-06-20
Add label to new chain addresses in Release notes

falvaradorodriguez:
To facilitate the reading of the Release notes, a section for automatic PRs of new chain addresses is added.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -7,6 +7,10 @@ changelog:
         labels:
           - dependencies
           - breaking_change
+          - add-new-address
+    - title: 🆕 New Chain Addresses
+      labels:
+        - add-new-address
     - title: 🛠 Breaking Changes
       labels:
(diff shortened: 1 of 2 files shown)


.github/release.yml
```diff
@@ -7,6 +7,10 @@ changelog:
         labels:
           - dependencies
           - breaking_change
+          - add-new-address
+    - title: 🆕 New Chain Addresses
```
Uxio0, line 11: I think this would be better below the rest of the changes, as this is not as relevant as the breaking changes

Uxio0 (changes requested):

.github/release.yml
falvaradorodriguez: Makes sense. Changed.

Uxio0 (approved):

PR #1209, opened 2024-07-09
Remove EthereumClientProvider

### LINKED ISSUE #1205: Remove EthereumClientProvider
This is not the best solution due to:
- Typing not working propertly
- Not the optimal way to implement a Singleton

Take a look at something like `web3.py auto`

Uxio0:
- That solution was very hacky, typing was not working correctly
- Create a new `get_auto_ethereum_client`
- Refactor code to use the new method
- Closes #1205

### DIFF
--- a/gnosis/util/util.py
+++ b/gnosis/util/util.py
@@ -1,12 +1,5 @@
 from typing import Any, Iterable, Sequence
 
-try:
-    from functools import cache
-except ImportError:
-    from functools import lru_cache
-
-    cache = lru_cache(maxsize=None)
-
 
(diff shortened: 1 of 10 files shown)


gnosis/util/util.py
```diff
@@ -1,12 +1,5 @@
 from typing import Any, Iterable, Sequence
 
-try:
-    from functools import cache
-except ImportError:
-    from functools import lru_cache
-
-    cache = lru_cache(maxsize=None)
```
falvaradorodriguez, line 8: good catch 👍

falvaradorodriguez (approved):

PR #1216, opened 2024-07-16
Make native balance optional for ERC20 `get_balances`

### LINKED ISSUE #1199: ERC20 `get_balances` should not return always the native balance
# What is needed?
The current implementation is adding always the Native token balance when we get the erc20 balance. 
https://github.com/safe-global/safe-eth-py/blob/72b7ed6daded3aa98cd536eb157551365f903a09/gnosis/eth/ethereum_client.py#L541

# Possible solutions
- Remove the Native token balance from the response of Erc20Manager.get_balance()
- Add an optional parameter setting if we client wants to include the Native token balance or not.

Uxio0:
- Add a new optional parameter so native balance is not alway returned
- Closes #1199

### DIFF
--- a/gnosis/eth/ethereum_client.py
+++ b/gnosis/eth/ethereum_client.py
@@ -517,13 +517,17 @@ class Erc20Manager(EthereumClientManager):
         )
 
     def get_balances(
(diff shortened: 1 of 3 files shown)


moisses89 (approved):
LGTM

PR #1308, opened 2024-08-29
Improve typing using `mypy`

### LINKED ISSUE #1272: Improve typing
Run mypy on the project and check if there are some easy typing problems that we can fix. It shouldn't be 100%, but I think with a little effort we can improve it and catch some potential bugs

This ticket should take at much 1 day of work

falvaradorodriguez:
Closes #1272 

Points of interest:

- The problems indicated by the mypy library are solved.
- mypy is configured in mypy.ini following the format of other Safe projects.
- Some exceptions need to be added:
  - Some `web3` lib related ignores and the abstract classes of the `safe` contracts. (`# type: ignore...` in specific lines)
  - Avoid checking `check_untyped_defs`. Checking it would involve rewriting a large part of the code. Right now only functions with type annotations are checked.
- Added `mypy` checking in CI and in the `run_tests.sh` script.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -29,7 +29,7 @@ jobs:
           path: ~/.cache/pre-commit
           key: pre-commit|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}
       - name: Install dependencies
-        run: pip install -U pre-commit
+        run: pip install -U pre-commit -r requirements-test.txt
       - name: Run pre-commit
         run: pre-commit run --all-files
 
@@ -67,8 +67,6 @@ jobs:
         pip install -r requirements-test.txt coveralls
       env:
         PIP_USE_MIRRORS: true
-    - name: Run mypy
-      run: mypy .
     - name: Run tests and coverage
       run: coverage run --source=$SOURCE_FOLDER -m pytest -W ignore::DeprecationWarning -rxXs --reruns 3
       env:
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -67,6 +67,8 @@ jobs:
         pip install -r requirements-test.txt coveralls
       env:
         PIP_USE_MIRRORS: true
+    - name: Run mypy
+      run: mypy .
     - name: Run tests and coverage
       run: coverage run --source=$SOURCE_FOLDER -m pytest -W ignore::DeprecationWarning -rxXs --reruns 3
       env:
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,5 +8,4 @@ py-evm==0.10.1b1
 pysha3>=1.0.0,<2.0.0; python_version<"3.9"
 requests==2.32.3
 safe-pysha3>=1.0.0; python_version>="3.9"
-types-requests==2.32.0.20240712
 web3==6.20.2
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,4 +8,5 @@ py-evm==0.10.1b1
 pysha3>=1.0.0,<2.0.0; python_version<"3.9"
 requests==2.32.3
 safe-pysha3>=1.0.0; python_version>="3.9"
+types-requests==2.32.0.20240712
 web3==6.20.2
--- a/safe_eth/eth/account_abstraction/bundler_client.py
+++ b/safe_eth/eth/account_abstraction/bundler_client.py
@@ -182,16 +182,15 @@ class BundlerClient:
         if (
             results
             and isinstance(results, list)
-            and results[0] is not None
+            and len(results) >= 2
             and isinstance(results[0], dict)
-            and results[1] is not None
             and isinstance(results[1], dict)
         ):
             return UserOperation.from_bundler_response(
                 user_operation_hash, results[0]
             ), self._parse_user_operation_receipt(results[1])
-        else:
-            return None
+
+        return None
 
     @lru_cache(maxsize=None)
     def supported_entry_points(self) -> List[ChecksumAddress]:
--- a/safe_eth/eth/account_abstraction/bundler_client.py
+++ b/safe_eth/eth/account_abstraction/bundler_client.py
@@ -210,9 +210,9 @@ class BundlerClient:
         result = self._do_request(payload)
         if result and isinstance(result, list):
             return [
-                ChecksumAddress(HexAddress(HexStr(x)))
-                for x in result
-                if isinstance(x, str)
+                ChecksumAddress(HexAddress(HexStr(address)))
+                for address in result
+                if isinstance(address, str)
             ]
-        else:
-            return []
+
+        return []
--- a/safe_eth/eth/account_abstraction/bundler_client.py
+++ b/safe_eth/eth/account_abstraction/bundler_client.py
@@ -2,13 +2,13 @@ import logging
 from functools import cache, lru_cache
 from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
 
-from eth_typing import ChecksumAddress, HexStr
+from eth_typing import ChecksumAddress, HexAddress, HexStr
 from hexbytes import HexBytes
 
 from safe_eth.util.http import prepare_http_session
 
 from .exceptions import BundlerClientConnectionException, BundlerClientResponseException
-from .user_operation import UserOperation
+from .user_operation import UserOperation, UserOperationV07
 from .user_operation_receipt import UserOperationReceipt
 
 logger = logging.getLogger(__name__)
@@ -33,7 +33,11 @@ class BundlerClient:
 
     def _do_request(
         self, payload: Union[Dict[Any, Any], Sequence[Dict[Any, Any]]]
-    ) -> Union[Optional[Union[Dict[str, Any]]], Optional[List[Dict[str, Any]]]]:
+    ) -> Union[
+        Optional[Union[Dict[str, Any]]],
+        Optional[List[Dict[str, Any]]],
+        Optional[List[str]],
+    ]:
         """
         :param payload: Allows simple request or a batch request
         :return: Result of the request
@@ -49,7 +53,7 @@ class BundlerClient:
 
         if not response.ok:
             raise BundlerClientConnectionException(
-                f"Error connecting to bundler {self.url} : {response.status_code} {response.content}"
+                f"Error connecting to bundler {self.url} : {response.status_code} {response.content!r}"
             )
 
         bundler_responses = response.json()
@@ -122,7 +126,7 @@ class BundlerClient:
     @lru_cache(maxsize=1024)
     def get_user_operation_by_hash(
         self, user_operation_hash: HexStr
-    ) -> Optional[UserOperation]:
+    ) -> Optional[Union[UserOperation, UserOperationV07]]:
         """
         https://docs.alchemy.com/reference/eth-getuseroperationbyhash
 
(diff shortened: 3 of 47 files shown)


safe_eth/eth/account_abstraction/bundler_client.py
```diff
@@ -173,13 +179,20 @@ def get_user_operation_and_receipt(
             self._get_user_operation_receipt_payload(user_operation_hash, request_id=2),
         ]
         result = self._do_request(payload)
-        if not (result and result[0]):
+        if (
```
moisses89, line 182: Why was necessary invert the condition?

safe_eth/eth/account_abstraction/bundler_client.py
```diff
@@ -195,4 +208,12 @@ def supported_entry_points(self) -> List[ChecksumAddress]:
...
         }
-        return self._do_request(payload)
+        result = self._do_request(payload)
+        if result and isinstance(result, list):
+            return [
+                ChecksumAddress(HexAddress(HexStr(x)))
+                for x in result
+                if isinstance(x, str)
+            ]
+        else:
```
moisses89, line 218: We don't need the else, right? I mean is the default return.

safe_eth/eth/account_abstraction/bundler_client.py
falvaradorodriguez: Just for simplicity, as the `_do_request` function returns different types, mypy forces you to check the type of each element that is accessed afterwards. This way I feel it is clearer.

safe_eth/eth/account_abstraction/bundler_client.py
falvaradorodriguez: We have discussed this previously, and we commented on maintaining it

requirements.txt
```diff
@@ -8,4 +8,5 @@ py-evm==0.10.1b1
 pysha3>=1.0.0,<2.0.0; python_version<"3.9"
 requests==2.32.3
 safe-pysha3>=1.0.0; python_version>="3.9"
+types-requests==2.32.0.20240712
```
Uxio0, line 11: I feel this should only be required for `dev` and not for production, am I right?

safe_eth/eth/account_abstraction/bundler_client.py
```diff
@@ -172,14 +178,21 @@ def get_user_operation_and_receipt(
...
+            and isinstance(results, list)
+            and results[0] is not None
+            and isinstance(results[0], dict)
+            and results[1] is not None
+            and isinstance(results[1], dict)
+        ):
+            return UserOperation.from_bundler_response(
+                user_operation_hash, results[0]
+            ), self._parse_user_operation_receipt(results[1])
+        else:
```
Uxio0, line 193: `else` is not required after `return`

safe_eth/eth/account_abstraction/bundler_client.py
```diff
@@ -172,14 +178,21 @@ def get_user_operation_and_receipt(
...
             self._get_user_operation_receipt_payload(user_operation_hash, request_id=2),
         ]
-        result = self._do_request(payload)
-        if not (result and result[0]):
+        results = self._do_request(payload)
+        if (
+            results
+            and isinstance(results, list)
+            and results[0] is not None
+            and isinstance(results[0], dict)
```
Uxio0, line 186:
```suggestion
            and isinstance(results[0], dict)
```

Shouldn't this be enough for `mypy`?

safe_eth/eth/account_abstraction/user_operation.py
```diff
@@ -204,7 +204,12 @@ def gas_fees(self) -> bytes:
 
     @property
     def paymaster_and_data(self) -> bytes:
-        if not self.paymaster:
+        if (
+            not self.paymaster
+            or not self.paymaster_verification_gas_limit
+            or not self.paymaster_post_op_gas_limit
+            or not self.paymaster_data
+        ):
```
Uxio0, line 212: Why are other parameters not related to this function checked here?

safe_eth/eth/clients/blockscout_client.py
```diff
@@ -135,8 +135,8 @@ class BlockscoutClient:
 
     def __init__(self, network: EthereumNetwork):
         self.network = network
-        self.grahpql_url = self.NETWORK_WITH_URL.get(network)
-        if self.grahpql_url is None:
+        self.grahpql_url = self.NETWORK_WITH_URL.get(network) or ""
+        if not self.grahpql_url:
```
Uxio0, line 139:
```suggestion
        self.grahpql_url: Optional[str] = self.NETWORK_WITH_URL.get(network)
        if self.grahpql_url is None:
```

Uxio0: Isn't this working?

Uxio0 (commented):
Really good job

Uxio0:
Mypy check should be added to `.pre-commit-config.yaml`

requirements.txt
falvaradorodriguez: Added in requirements-test ([here](https://github.com/safe-global/safe-eth-py/pull/1308/commits/1d71d495526362616735c2fdb81c507054333525)), to use it in [CI validation](https://github.com/safe-global/safe-eth-py/blob/68938a79ebc7ffdd0f4b33cc24ee19e05d436368/.github/workflows/python.yml#L67).

safe_eth/eth/account_abstraction/bundler_client.py
falvaradorodriguez: Fixed [here](https://github.com/safe-global/safe-eth-py/pull/1308/commits/1d71d495526362616735c2fdb81c507054333525)

safe_eth/eth/account_abstraction/bundler_client.py
falvaradorodriguez: It's enough, but you need to check if results have at least two elements to avoid `list index out of range`. Updated [here](https://github.com/safe-global/safe-eth-py/pull/1308/commits/1d71d495526362616735c2fdb81c507054333525).

safe_eth/eth/account_abstraction/user_operation.py
falvaradorodriguez: All of "if" params are being used in this [return statement](https://github.com/safe-global/safe-eth-py/blob/c0609e98a31daac5872c5151dfde7cd2ea7fb6c2/safe_eth/eth/account_abstraction/user_operation.py#L214), isn't it?

falvaradorodriguez:
> Mypy check should be added to `.pre-commit-config.yaml`

@Uxio0 Added [here](https://github.com/safe-global/safe-eth-py/pull/1308/commits/c0609e98a31daac5872c5151dfde7cd2ea7fb6c2).

I put this in a separate commit in case you prefer to remove it, as the way to run the hook is a bit complex. It runs in a separate `env` and needs to indicate what dependencies it needs, also it needs to replicate the configuration, as it doesn't take into account the `mypy.ini` file. In my opinion I think it would be more maintainable to run it separately with `mypy.ini` and also in the CI. 

What do you think?

safe_eth/eth/account_abstraction/bundler_client.py
Uxio0: Makes sense

safe_eth/eth/account_abstraction/bundler_client.py
Uxio0: Some tools recommend to remove the `else` after a `return` (like pylint if I'm not wrong). I would remove it. Also, please don't use `x` for a list comprehension, and use the name of the variable you expect, maybe `address` in this case

safe_eth/eth/account_abstraction/bundler_client.py
Uxio0: Nice

safe_eth/eth/account_abstraction/user_operation.py
Uxio0: You are totally right, I was not looking at the right function 🤦🏻

safe_eth/eth/clients/blockscout_client.py
Uxio0: Ok, then `self.grahpql_url = self.NETWORK_WITH_URL.get(network, "")` would be better 😉

Uxio0:
> > Mypy check should be added to `.pre-commit-config.yaml`
> 
> @Uxio0 Added [here](https://github.com/safe-global/safe-eth-py/pull/1308/commits/9326eb983452828be25563ed3f616a5e364029b8).
> 
> I put this in a separate commit in case you prefer to remove it, as the way to run the hook is a bit complex. It runs in a separate `env` and needs to indicate what dependencies it needs, also it needs to replicate the configuration, as it doesn't take into account the `mypy.ini` file. In my opinion I think it would be more maintainable to run it separately (or with [`run_tests.sh`](https://github.com/safe-global/safe-eth-py/blob/065af53ee8be26f12a9d53248626e1e7371fa189/run_tests.sh#L4)) with `mypy.ini` and also in the [CI](https://github.com/safe-global/safe-eth-py/blob/6cc6687d6afa8a04fa4e658c9267f3dd72eb3d3b/.github/workflows/python.yml#L70).
> 
> What do you think?

Ok, we can do it in a separate PR, makes sense

Uxio0 (approved):

safe_eth/eth/clients/blockscout_client.py
falvaradorodriguez: Yes, I already updated this in the commit that I mentioned above :)

safe_eth/eth/account_abstraction/bundler_client.py
falvaradorodriguez: Ok, makes sense, updated [here](https://github.com/safe-global/safe-eth-py/pull/1308/commits/d5124b9ba7afdab46bf2237d16f63c7072d97114).

.github/workflows/python.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           path: ~/.cache/pre-commit
           key: pre-commit|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}
       - name: Install dependencies
-        run: pip install -U pre-commit
+        run: pip install -U pre-commit -r requirements-test.txt
```
falvaradorodriguez, line 32: Necessary to be able to run the mypy hook in the linting check step.

falvaradorodriguez:
> > > Mypy check should be added to `.pre-commit-config.yaml`
> > 
> > 
> > @Uxio0 Added [here](https://github.com/safe-global/safe-eth-py/pull/1308/commits/9326eb983452828be25563ed3f616a5e364029b8).
> > I put this in a separate commit in case you prefer to remove it, as the way to run the hook is a bit complex. It runs in a separate `env` and needs to indicate what dependencies it needs, also it needs to replicate the configuration, as it doesn't take into account the `mypy.ini` file. In my opinion I think it would be more maintainable to run it separately (or with [`run_tests.sh`](https://github.com/safe-global/safe-eth-py/blob/065af53ee8be26f12a9d53248626e1e7371fa189/run_tests.sh#L4)) with `mypy.ini` and also in the [CI](https://github.com/safe-global/safe-eth-py/blob/6cc6687d6afa8a04fa4e658c9267f3dd72eb3d3b/.github/workflows/python.yml#L70).
> > What do you think?
> 
> Ok, we can do it in a separate PR, makes sense

@Uxio0 In the end I opted to put `language: system` in the mypy hook configuration in pre-commit ([here](https://github.com/safe-global/safe-eth-py/pull/1308/commits/7adb85c5b166d41a0decc6ae7ff4abd471ae014d)) and remove the `run_tests` checks and the `test-app` IC ([here](https://github.com/safe-global/safe-eth-py/pull/1308/commits/3f0e6819e67db1430d2f6d9724b0afab9bf8feab)). 

The check will be run locally every time a commit is made and in the CI in the linting check phase.

PR #1380, opened 2024-10-10
Support multiple types of proxies

Uxio0:
- Until now, Safe only supported Safe Proxies
- Now ERC-1167 and ERC-1967 proxies are also supported

### DIFF
--- a/safe_eth/eth/proxies/minimal_proxy.py
+++ b/safe_eth/eth/proxies/minimal_proxy.py
@@ -1,0 +1,59 @@
+from functools import cache
+from typing import Optional
+
+from eth_typing import ChecksumAddress
+from hexbytes import HexBytes
+from web3.types import BlockIdentifier
+
+from ..constants import NULL_ADDRESS
+from ..utils import fast_to_checksum_address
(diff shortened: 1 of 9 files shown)


safe_eth/eth/proxies/minimal_proxy.py
```diff
@@ -0,0 +1,59 @@
...
+        :return: Expected code for a given `contract_address`
+        """
+        return (
+            HexBytes("363d3d373d3d3d363d73")
+            + HexBytes(implementation_address)
+            + HexBytes("5af43d82803e903d91602b57fd5bf3")
+        )
+
+    @cache
+    def get_singleton_address(
```
moisses89, line 47: Maybe we can or we would like to reuse this to other proxy implementations, could we call instead get_singleton_address call it as get_implementation_address?

moisses89 (approved):

safe_eth/eth/proxies/minimal_proxy.py
Uxio0: You are right, done

PR #1399, opened 2024-10-25
Fix CLA signatures Github action

luarx:
It was broken as the [last CLA change](https://github.com/safe-global/safe-eth-py/commit/61d00aefa3959a818a04ee1fc6f1da319ef0aa97) needs more modifications

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -21,12 +21,14 @@ jobs:
           path-to-signatures: 'signatures/version1/cla.json'
           path-to-document: 'https://safe.global/cla'
           # branch should not be protected
-          branch: 'cla-signatures'
+          branch: 'main'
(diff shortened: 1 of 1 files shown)


.github/workflows/cla.yml
```diff
@@ -21,12 +21,14 @@ jobs:
           path-to-signatures: 'signatures/version1/cla.json'
           path-to-document: 'https://safe.global/cla'
           # branch should not be protected
-          branch: 'cla-signatures'
+          branch: 'main'
```
Uxio0, line 24: Why this change? We don't want the signatures to be stored in our main branch

luarx: They will not be stored in the safe-eth-py repo, but in a centralised repo as we are doing with other repos that is specified in the following line ` remote-repository-name: 'cla-signatures'` (https://github.com/safe-global/cla-signatures)

Uxio0 (approved):
Makes sense, I didn't see the repo change! Sorry!

Issue #1398, opened 2024-10-25
Support Etherscan V2 API


Uxio0:
More info: https://docs.etherscan.io/etherscan-v2

Olowojaye:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

PR #1407, opened 2024-10-31
Add new client for Etherscan API v2

falvaradorodriguez:
- Closes: #1398

More info about the migration: https://docs.etherscan.io/etherscan-v2/getting-started/v2-quickstart

### DIFF
--- a/safe_eth/eth/clients/etherscan_client.py
+++ b/safe_eth/eth/clients/etherscan_client.py
@@ -319,8 +319,8 @@ class EtherscanClient:
         self.http_session.headers = self.HTTP_HEADERS
         self.request_timeout = request_timeout
 
-    def build_url(self, path: str):
-        url = urljoin(self.base_api_url, f"api?{path}")
+    def build_url(self, query: str):
+        url = urljoin(self.base_api_url, f"api?{query}")
         if self.api_key:
             url += f"&apikey={self.api_key}"
         return url
--- a/safe_eth/eth/clients/etherscan_client.py
+++ b/safe_eth/eth/clients/etherscan_client.py
@@ -320,7 +320,7 @@ class EtherscanClient:
         self.request_timeout = request_timeout
 
     def build_url(self, path: str):
-        url = urljoin(self.base_api_url, path)
+        url = urljoin(self.base_api_url, f"api?{path}")
         if self.api_key:
             url += f"&apikey={self.api_key}"
         return url
@@ -385,7 +385,7 @@ class EtherscanClient:
         :return:
         """
         url = self.build_url(
-            f"api?module=contract&action=getsourcecode&address={contract_address}"
+            f"module=contract&action=getsourcecode&address={contract_address}"
         )
         response = self._retry_request(url, retry=retry)  # Returns a list
         if response and isinstance(response, list):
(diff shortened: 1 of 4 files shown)


safe_eth/eth/clients/etherscan_client_v2.py
```diff
@@ -0,0 +1,56 @@
...
+        request_timeout: int = int(
+            os.environ.get("ETHERSCAN_CLIENT_REQUEST_TIMEOUT", 10)
+        ),
+    ):
+        super().__init__(EthereumNetwork.MAINNET, api_key, request_timeout)
+        self.network = network
+        self.base_api_url = self.BASE_API_V2_URL
+
+    @classmethod
+    def _get_supported_networks(cls) -> List[Dict[str, Any]]:
```
Uxio0, line 27: I think this should be cached, and also it shouldn't be a private method

safe_eth/eth/clients/etherscan_client_v2.py
```diff
@@ -0,0 +1,56 @@
...
+
+        :param network: The Ethereum network to check.
+        :return: True if the network is supported; False otherwise.
+        """
+        supported_networks = cls._get_supported_networks()
+        return any(
+            item.get("chainid") == str(network.value) for item in supported_networks
+        )
+
+    def build_url(self, path: str) -> str:
```
Uxio0, line 52: This method should be at the top of the class, just for clarity

Uxio0 (changes requested):

safe_eth/eth/clients/etherscan_client.py
```diff
@@ -320,7 +320,7 @@ def __init__(
         self.request_timeout = request_timeout
 
     def build_url(self, path: str):
-        url = urljoin(self.base_api_url, path)
+        url = urljoin(self.base_api_url, f"api?{path}")
```
moisses89, line 323: Why api is out of the path?

safe_eth/eth/clients/etherscan_client_v2.py
```diff
@@ -0,0 +1,56 @@
...
+        url = urljoin(self.base_api_url, f"v2/api?chainid={self.network.value}&{path}")
```
moisses89, line 53: `v2/api` should be path?

moisses89 (approved):

safe_eth/eth/clients/etherscan_client_v2.py
falvaradorodriguez: Done [here](https://github.com/safe-global/safe-eth-py/pull/1407/commits/9b8049300f80f04aa75dd4b1f643d0735ab3d0ed)

safe_eth/eth/clients/etherscan_client_v2.py
falvaradorodriguez: IMO this method will be called a few times and may even add a string along the way. It should have no problems without being cached.

It is updated as public method [here](https://github.com/safe-global/safe-eth-py/pull/1407/commits/9b8049300f80f04aa75dd4b1f643d0735ab3d0ed)

safe_eth/eth/clients/etherscan_client.py
falvaradorodriguez: Maybe the problem is the name of the parameter. I updated it as `query` [here](https://github.com/safe-global/safe-eth-py/pull/1407/commits/9b8049300f80f04aa75dd4b1f643d0735ab3d0ed)

safe_eth/eth/clients/etherscan_client_v2.py
falvaradorodriguez: Yes, it's the path. I already updated the name of the parameter.

Uxio0 (approved):

PR #1410, opened 2024-10-31
Add missing deployment addresses

falvaradorodriguez:
- Related with: https://github.com/safe-global/safe-platform-requests/issues/11


Uxio0 (approved):

Uxio0:
@falvaradorodriguez we need to be careful when deploying this change, as it will get back our indexer to the block number

PR #1439, opened 2024-11-22
Check batch requests return same number of results

Uxio0:
- It could be possible that a batch request returns less results than the number of payloads
- We should catch and log those if that's the case


falvaradorodriguez (approved):

falvaradorodriguez:
It seems that the tests need to be adjusted.

moisses89 (approved):

PR #1505, opened 2024-12-30
Add async methods to source contracts clients

moisses89:
# Description
- Add asynchronous methods of SourcifyClient, EtherscanClientV2 and BlockScoutClient. 
- EtherscanClient V1 is not included.

# Considerations
It was necessary extend the class with different method names because `mypy` does not let override functions with async and not asynchronous applications could have troubles  generating `aiohttp` session.

### DIFF
--- a/safe_eth/eth/clients/etherscan_client.py
+++ b/safe_eth/eth/clients/etherscan_client.py
@@ -366,6 +366,21 @@ class EtherscanClient:
                 return ContractMetadata(contract_name, contract_abi, False)
         return None
 
+    def process_response(self, response):
+        if response and isinstance(response, list):
+            result = response[0]
+            abi_str = result.get("ABI")
+
+            if isinstance(abi_str, str) and abi_str.startswith("["):
+                try:
+                    result["ABI"] = json.loads(abi_str)
+                except json.JSONDecodeError:
+                    result["ABI"] = None  # Handle the case where JSON decoding fails
+            else:
+                result["ABI"] = None
+
+            return result
+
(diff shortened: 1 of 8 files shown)


safe_eth/eth/clients/etherscan_client.py
```diff
@@ -366,6 +366,21 @@ def get_contract_metadata(
                 return ContractMetadata(contract_name, contract_abi, False)
         return None
 
+    def process_response(self, response):
```
falvaradorodriguez, line 369: This name seems a bit generic to me. It does not process all types of Etherescan api responses, it processes a specific type. 

Should it be a private method?

safe_eth/eth/clients/etherscan_client_v2.py
```diff
@@ -78,3 +84,81 @@ def is_supported_network(cls, network: EthereumNetwork) -> bool:
...
+        response = await self._async_do_request(url)  # Returns a list
+        return self.process_response(response)
+
+    async def async_get_contract_metadata(
+        self, contract_address: str
+    ) -> Optional[ContractMetadata]:
+        contract_source_code = await self.async_get_contract_source_code(
+            contract_address
+        )
+        if contract_source_code:
```
falvaradorodriguez, line 145: Would it make sense to add this processing in a private method of the superclass and call it here?

Same for the other similar cases in other clients.

WDYT?

moisses89: Done @falvaradorodriguez

falvaradorodriguez (approved):

PR #1536, opened 2025-01-29
Bump safe-deployments to last main version

moisses89:


moisses89:
@Uxio0 the change on the order is because this change https://github.com/safe-global/safe-eth-py/pull/1489/files

Uxio0 (approved):

PR #1546, opened 2025-02-05
Remove deprecated serializers

falvaradorodriguez:
- Related with https://github.com/safe-global/safe-transaction-service/issues/2395


Uxio0 (approved):

Uxio0 (changes requested):
Serializers should be ready for people who want to use them. MultisigTransaction is lacking fields like base_gas, etc.

falvaradorodriguez:
> Serializers should be ready for people who want to use them. MultisigTransaction is lacking fields like base_gas, etc.

Ok, makes sense. I close this PR then.

PR #1557, opened 2025-02-12
Fix test_send_unsigned_transaction

moisses89:
RPC is checking first if the account have funds, so the chainId is required is not being raised.

### DIFF
--- a/safe_eth/eth/tests/clients/test_blockscout_client.py
+++ b/safe_eth/eth/tests/clients/test_blockscout_client.py
@@ -1,7 +1,5 @@
 import unittest
 
-from django.test import TestCase
-
 import pytest
 
 from ... import EthereumNetwork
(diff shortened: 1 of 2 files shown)


safe_eth/eth/tests/clients/test_blockscout_client.py
```diff
@@ -10,7 +8,7 @@
 from .mocks import safe_proxy_abi_mock, sourcify_safe_metadata
 
 
-class TestBlockscoutClient(TestCase):
+class TestBlockscoutClient(unittest.TestCase):
```
moisses89, line 11: I think is not necessary Django support for this test because there is not any Django model involved.
Extending Django TestCase require a database connection and is not necessary for this test.

Uxio0: Agree, are more tests inheriting from Django TestCase?

Uxio0 (approved):

safe_eth/eth/tests/clients/test_blockscout_client.py
moisses89: I´ll keep an eye on it.

PR #1589, opened 2025-03-12
Fix tx-service api url builder

### LINKED ISSUE #1575: TransactionServiceAPI truncates its base_url parameter when making API requests
**Describe the bug**
If you set a base_url for `safe_eth.safe.api.TransactionServiceApi` then any part of it past the domain name will be chopped off before creating the path for making API HTTP requests.

E.g. if you set base_url="localhost:8000" and base_url="localhost:8000/txs" you will get the same result, as in the second case the "txs" part will be lost.

This is ultimately due to passing API paths with a leading slash which causes urljoin to truncate the base_url.
See the following behaviour:
```
(Pdb) urljoin("http://a.com/b/", "/c")
'http://a.com/c'
(Pdb) urljoin("http://a.com/b/", "c")
'http://a.com/b/c'
(Pdb) urljoin("http://a.com/b", "c")
'http://a.com/c'
```

So because your code base has snippets such as transaction_service_api.py:180:
```
        response = self._get_request(
            f"/api/v1/multisig-transactions/{safe_tx_hash_str}/"
        )
```
the base_url will always get truncated.

**To Reproduce**
1. import safe_eth.safe.api.TransactionServiceApi
2. t = safe_eth.safe.api.TransactionServiceApi(..., base_url=localhost:8000/txs")
3. make some call using `t`

notably the safe-infrastructure repo has a docker-compose.yml which hosts the tx service at localhost:8000/txs, so this bug makes it difficult to use safe-eth-py with that docker setup.

**Expected behavior**
The base url should not be truncated. Any path after the domain name in the base url should end up in the http request path.

**Additional context**
n/a

falvaradorodriguez:
- Closes #1575

### DIFF
--- a/safe_eth/util/http.py
+++ b/safe_eth/util/http.py
@@ -49,10 +49,7 @@ def build_full_url(base_url: str, url: str) -> str:
     if not base_url.startswith(("http://", "https://")):
         base_url = f"http://{base_url}/"
 
-    if base_url.endswith("/"):
-        base_url = base_url.rstrip("/")
-
-    if url.startswith("/"):
-        url = url.lstrip("/")
+    base_url = base_url.rstrip("/")
+    url = url.lstrip("/")
 
     return urljoin(f"{base_url}/", url)
--- a/safe_eth/util/http.py
+++ b/safe_eth/util/http.py
@@ -1,4 +1,5 @@
 from typing import Union
+from urllib.parse import urljoin
 
 import requests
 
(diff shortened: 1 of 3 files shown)


safe_eth/util/http.py
```diff
@@ -32,3 +33,26 @@ def prepare_http_session(
...
+    :return: The complete URL formed by combining base_url and url.
+    """
+    if not base_url.startswith(("http://", "https://")):
+        base_url = f"http://{base_url}/"
+
+    if base_url.endswith("/"):
+        base_url = base_url.rstrip("/")
+
+    if url.startswith("/"):
+        url = url.lstrip("/")
```
Uxio0, line 56:
```suggestion
    base_url = base_url.rstrip("/")
    url = url.lstrip("/")
```

Uxio0 (approved):

PR #1603, opened 2025-03-25
Add default safe deployment addresses

moisses89:


scripts/generators/generate_safe_deployments.py
```diff
@@ -37,14 +40,14 @@ def get_safe_deployments(*args, **kwds):
         clean_resources()
 
 
-def get_network_addresses(safe_deployment: Dict) -> Dict[str, List[str]]:
+def get_network_addresses_by_chain(safe_deployment: Dict) -> Dict[str, List[str]]:
     """
     Convert canonical and non-canonical from safe-deployments json to the corresponding addresses.
 
     :param safe_deployment:
     :return:
```
Uxio0, line 48: Please add more info here

scripts/generators/generate_safe_deployments.py
```diff
@@ -37,14 +40,14 @@ def get_safe_deployments(*args, **kwds):
         clean_resources()
 
 
-def get_network_addresses(safe_deployment: Dict) -> Dict[str, List[str]]:
+def get_network_addresses_by_chain(safe_deployment: Dict) -> Dict[str, List[str]]:
```
Uxio0, line 43: Typing. Chain should be an int

scripts/generators/generate_safe_deployments.py
```diff
@@ -55,9 +58,25 @@ def get_network_addresses(safe_deployment: Dict) -> Dict[str, List[str]]:
...
+
+    return network_addresses_by_chain
+
+
+def get_default_network_addresses(deployments: Dict) -> List[str]:
+    """
+    Get addresses from deployments dict.
+
+    :param deployments: deployments dict on safe_deployments
+    :return:
```
Uxio0, line 71: Doc

Uxio0 (approved):

scripts/generators/generate_safe_deployments.py
moisses89: Actually chain is string in the dict, no ints. 
Agree with change to int, but this would be a breaking change for the library. 
Do you want to update it to int in this PR?

PR #1704, opened 2025-06-06
Use new 1271 validation method for Safes 1.3.0/1.4.1

Uxio0:


moisses89 (approved):

Uxio0:
No needed for this version, it would break a lot and currently the other method is still available and even used by `checkSignatures`

PR #1748, opened 2025-06-10
Fix: Ensure HexV2Field properly handles DB and Python values

LucasCambon:
### Description:

This PR improves HexV2Field by implementing the full Django value conversion interface:

- `from_db_value()`: ensures `bytea` values loaded from PostgreSQL are returned as `HexBytes`, instead of `memoryview`.
- `to_python()`: normalizes inputs like `bytes`, `HexBytes`, and `"0x..."` strings to `HexBytes`.
- `get_prep_value()`: ensures only raw `bytes` are sent to the DB, avoiding errors from passing a hex string directly.

This eliminates the need for consumers to subclass the field or manually convert values when reading/writing to the DB.

Fixes https://github.com/safe-global/safe-eth-py/issues/1749

### DIFF
--- a/safe_eth/eth/django/models.py
+++ b/safe_eth/eth/django/models.py
@@ -196,6 +196,26 @@ class HexV2Field(models.BinaryField):
         defaults.update(kwargs)
         return super().formfield(**defaults)
 
+    def from_db_value(self, value, expression, connection):
+        if value:
+            return HexBytes(bytes(value))
+        return None
+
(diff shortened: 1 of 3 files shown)


moisses89:
Thanks for your research but we think this is not necessary, the memoryview was returned for older versions of psycopg.
More information [here](https://github.com/safe-global/safe-eth-py/issues/1749#issuecomment-2990919932)

PR #1752, opened 2025-06-11
Add api key to Safe services API and update base URLs

### LINKED ISSUE #1692: feat: update tx-service to API Gateway
# What is needed?
Now that we have an API Gateway, that should be used instead of the direct access to the TX Service. Also it should be possible to add an API Key to the project configuration so calls can be authenticated to the new Gateway.

# Background
It should be possible to use `safe-eth-py` through the new API Gateway
It should be possible to add the API Key so requests can be correctly authenticated
If the TX Service URL is currently configurable for externally hosted TX Services it should be possible to skip the API Key property to allow this services to use the `safe-eth-py` as usual.

# Related issues
[Paste here the related links for the issues on the clients/safe project if applicable](https://github.com/safe-global/safe-core-sdk/issues/1214)

falvaradorodriguez:
- Closes #1692


safe_eth/safe/api/base_api.py
```diff
@@ -16,52 +16,74 @@ class SafeAPIException(Exception):
...
         :param ethereum_client:
         :param base_url: If a custom transaction service is used
+        :param api_key: Api key of authenticated Safe services
         :raises: EthereumNetworkNotSupported
         """
         self.network = network
         self.ethereum_client = ethereum_client
-        self.base_url = base_url or self.URL_BY_NETWORK.get(network, "")
-        if not self.base_url:
+        base_url = base_url or self.get_url_by_network(network)
```
moisses89, line 36: If the purpose of `get_url_by_network` is just fill base_url then I think that the method should be private.

safe_eth/safe/api/base_api.py
```diff
@@ -16,52 +16,74 @@ class SafeAPIException(Exception):
...
+        if not base_url:
             raise EthereumNetworkNotSupported(network)
+        self.base_url = base_url
```
moisses89, line 39:
```suggestion
        self.base_url = base_url or self.get_url_by_network(network)
```
Also I think this exception `raise EthereumNetworkNotSupported(network)`, should be raised by `get_url_by_network`.

safe_eth/safe/api/base_api.py
```diff
@@ -16,52 +16,74 @@ class SafeAPIException(Exception):
...
     def _delete_request(self, url: str, payload: Dict) -> requests.Response:
         full_url = build_full_url(self.base_url, url)
+        headers = {"Content-Type": "application/json"}
+        if self.api_key:
+            headers["Authorization"] = f"Bearer {self.api_key}"
         return self.http_session.delete(
             full_url,
             json=payload,
-            headers={"Content-type": "application/json"},
+            headers=headers,
```
moisses89, line 87: I think that headers could be refactored to get_request_headers and just do` headers=self._get_request_headers()`

safe_eth/safe/tests/api/test_transaction_service_api.py
```diff
@@ -36,11 +49,37 @@ def test_constructor(self):
         ethereum_network = EthereumNetwork.SEPOLIA
         base_url = "https://safe.global"
```
moisses89, line 50: Because we have a previous API url without authentication, could you try to configure any network and do a get to any endpoint?

safe_eth/safe/api/base_api.py
falvaradorodriguez: Done https://github.com/safe-global/safe-eth-py/pull/1752/commits/d582274b25482061d5b747b1ad79d848a9b9978b

safe_eth/safe/api/base_api.py
falvaradorodriguez: Since this method must be implemented in the child class and not in this one, I have added a new method that calls the abstract method that implements the url per network and does the exception handling. https://github.com/safe-global/safe-eth-py/pull/1752/commits/d582274b25482061d5b747b1ad79d848a9b9978b

PR #1969, opened 2025-09-22
Feat/support v1.5.0

akshay-ap:
Changes in PR:
- Add v1.5.0 related files
- Update code to use v1.5.0 addresses

TODO: 
- [ ] Fix tests
- [ ] Update safe deployments `SAFE_DEPLOYMENTS_VERSION` (add network `1337` for v1.5.0)


safe_eth/safe/safe_deployments.py
moisses89, line 1: I think this can be handled in a different PR because as far as I know is not being used for Safe class. What do you think?

falvaradorodriguez:
Since this PR was a little behind main and changes have been made to safe_signature, I have decided to start a new branch taking into account the changes made here.

The PR is replaced by: https://github.com/safe-global/safe-eth-py/pull/2179

PR #2179, opened 2025-11-21
feat: Support safe contracts v1.5.0

### LINKED ISSUE #1930: [Tests] Adapt test for 1.5.0 compatiblity
# What is needed?

# Background
More information about the feature needed

# Related issues
Paste here the related links for the issues on the clients/safe project if applicable

### LINKED ISSUE #2124: Supporting Safe address of 1.5.0
Is the this library can support Safe contract 1.5.0?

My organization currently operates an Arbitrum Orbit chain and is trying to run its own safe infrastructure to facilitate multisig wallet support on this chain.

Recently, I completed to deploy a contract using safe-smart-account and registered it in the safe-deployments repository. I understood that in order to run the safe transaction service properly, I needed to add the chain to safe-eth-py.

However, the safe-smart-account guide indicated that the current stable contract version was 1.5.0, so I deployed that version. However, when I tried to register an issue to safe-eth-py repository, I was only able to choose between two versions: 1.3 and 1.4.1.

falvaradorodriguez:
- Closes #2124
- Closes #1930

To take into account:

- The deprecated function isValidSignature(bytes, bytes) is no longer available in version 1.5.0. (https://github.com/safe-fndn/safe-smart-account/commit/2fb54f1839b2c689f142e0011cc9ad9ec043aaf6) ~~**This implies that contracts <1.3.0 will no longer be supported** if we follow the current approach of using the same function to validate all signatures. Otherwise, we would have to manage the version before each signature validation.~~ The latest commit adds compatibility for versions that do not yet implement the deprecated function. If it is not possible to verify validity with the new function signature, the previous function signature is used.

### DIFF
--- a/safe_eth/safe/safe.py
+++ b/safe_eth/safe/safe.py
@@ -69,7 +69,7 @@ class SafeInfo:
     owners: List[ChecksumAddress]
     threshold: int
     version: str
-    module_guard: ChecksumAddress = NULL_ADDRESS
+    module_guard: ChecksumAddress
 
 
 class Safe(SafeCreator, ContractBase, metaclass=ABCMeta):


safe_eth/safe/tests/test_safe_signature.py
```diff
@@ -80,6 +85,42 @@ def build_eip1271_signature_for_message(test_case: "SafeTestCaseMixin"):
 
 
 def setup_safe_contract_signature_case(test_case: "SafeTestCaseMixin"):
```
akshay-ap: shouldn't this be `setup_safe_contract_signature_case_v_1_3_0`?

safe_eth/safe/safe.py
```diff
@@ -1163,12 +1167,14 @@ def get_contract_fn(self) -> Callable[[Web3, Optional[ChecksumAddress]], Contrac
         return get_safe_V1_3_0_contract
 
 
-class SafeV141(Safe):
-    def get_version(self) -> str:
-        return "1.4.1"
+class SafeLegacyAdapter(Safe, ABC):
```
Uxio0: Add comments to explain this class

falvaradorodriguez:
> I think [mock_bundler.py](https://github.com/safe-global/safe-eth-py/blob/5194ad6fbb9d5b4f807934521897bf0e4e59883a/safe_eth/eth/tests/mocks/mock_bundler.py#L29) should be also updated with v1.5.0 factory address.

Updated [here](https://github.com/safe-global/safe-eth-py/pull/2179/commits/0429037378b8d109cd3047dd4499293d0d017484). Could you review? @akshay-ap

falvaradorodriguez:
> I think [mock_bundler.py](https://github.com/safe-global/safe-eth-py/blob/5194ad6fbb9d5b4f807934521897bf0e4e59883a/safe_eth/eth/tests/mocks/mock_bundler.py#L29) should be also updated with v1.5.0 factory address.

Thanks to catch it! Fixed [here](https://github.com/safe-global/safe-eth-py/pull/2179/commits/81b55eaf7de62db9cadcaf8dadfa333ed09cf39d)

safe_eth/safe/safe_signature.py
```diff
@@ -405,35 +406,82 @@ def signature_type(self) -> SafeSignatureType:
...
+        )
+
     def is_valid(
         self,
         ethereum_client: Optional[EthereumClient] = None,
         safe_address: Optional[str] = None,
     ) -> bool:
+        """
+        Validate the signature using the appropriate EIP-1271 path.
+        First tries the updated method (bytes32,bytes).
```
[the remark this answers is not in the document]
falvaradorodriguez: This would be fine if you always know in advance the form of the signature and the version of the safe with which it was created.

The way I see it, the signature may not be aware of this information and still work. As for performance for safes > 1.3.0, it will always be validated on the first attempt. 

Furthermore, adding the flag would be a breaking change, and future updates would be more costly. Right now, if we decide to deprecate safes <1.3.0, we would simply delete the logic and the signature of the function would be the same.

safe_eth/safe/safe.py
```diff
@@ -86,7 +89,7 @@ class Safe(SafeCreator, ContractBase, metaclass=ABCMeta):
     SAFE_MESSAGE_TYPEHASH = bytes.fromhex(
         "60b3cbf8b4a223d68d641b3b6ddf9a298e7f33710cf3d3a9d1146b5a6150fbca"
     )
-    _DEFAULT_VERSION = "1.4.1"
+    _DEFAULT_VERSION = "1.5.0"
```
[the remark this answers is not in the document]
falvaradorodriguez: Thank you very much for mentioning it and for the indications and examples. Good point! I've updated the Safe class [here](https://github.com/safe-global/safe-eth-py/pull/2179/commits/ae7ce11169853f12501eea8cc7d02e3df087dc07). Please, check if everything it's ok.

safe_eth/safe/safe_signature.py
```diff
@@ -405,35 +406,82 @@ def signature_type(self) -> SafeSignatureType:
 
 
 class SafeSignatureContract(SafeSignatureContractMixin, SafeSignature):
+    def _check_eip1271(
+        self,
+        ethereum_client: EthereumClient,
+        handler_getter,
```
Uxio0: typing `Callable[.......]`

safe_eth/safe/safe_signature.py
```diff
@@ -405,35 +406,82 @@ def signature_type(self) -> SafeSignatureType:
...
-                self.EIP1271_MAGIC_VALUE,
-                self.EIP1271_MAGIC_VALUE_UPDATED,
-            )
-        except (Web3Exception, DecodingError, Web3ValueError, Web3RPCError):
-            # Error using `pending` block identifier or contract does not exist
-            logger.warning(
-                "Cannot check EIP1271 signature from contract %s", self.owner
-            )
+
+        for handler_getter, function_signature, data in (
```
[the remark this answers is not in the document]
Uxio0: IMO `handler` is very generic

safe_eth/safe/safe_signature.py
```diff
@@ -488,33 +536,80 @@ def is_valid(
...
+        Attempt to validate an EIP-1271 signature using a specific CompatibilityFallbackHandler and function signature.
+
+        :param web3: AsyncWeb3 instance
+        :param handler_getter: A function that returns the appropriate CompatibilityFallbackHandler contract for the current Safe.
+        :param function_signature: The ABI function signature to call
+        :param data: The data or hash to be validated, depending on the Safe version.
+        :param signature: The contract signature payload to validate.
+        :return: True on successful validation; otherwise False.
+        """
+        handler = handler_getter(web3, self.owner)
```
Uxio0: Not liking duplication of cod, but also I cannot think of an easiest way to do it

Uxio0 (approved):
Well done. Some small comments

akshay-ap:
While testing the PR with Safe Transaction service I noticed this issue:

Existing v1.4.1 uses L2
https://github.com/safe-global/safe-eth-py/blob/8b30e03a9e6d0df273bc4f6e15238a65dbf58816/safe_eth/eth/contracts/abis/Safe_V1_4_1.json#L3 

While new v1.5.0: uses L1:

https://github.com/safe-global/safe-eth-py/blob/8b30e03a9e6d0df273bc4f6e15238a65dbf58816/safe_eth/eth/contracts/abis/Safe_V1_5_0.json#L3 


Safe transaction service assumes that the contract is `L2` which emits additional [SafeMultiSigTransaction](https://github.com/safe-fndn/safe-smart-account/blob/cdb2eb578dbdba4c3f10a47f9a2dd9580773e63a/contracts/SafeL2.sol#L33) event. Now, with `L1`, this event is not emitted and some tests are failing in safe transaction service. We should keep the version consistent. In this case use v1.5.0 L2 version.

Ref: https://github.com/safe-global/safe-transaction-service/blob/8b495ecc7da26307f45fe6d669a99cb615fd2b18/safe_transaction_service/history/tests/test_safe_events_indexer.py#L269

akshay-ap:
LGTM! Great work!

PR #2426, opened 2026-04-30
chore: improve existing BlockscoutClient detection

dasanra:
## Summary

Currently the script adds a new line if the explorer is added with or without trailing slash. We detect if a client is already added and in that case we skip it.

## Other considerations

I was considering updating Client URL if it was detected for an existing network and different links.


falvaradorodriguez (approved):

falvaradorodriguez:
Please @dasanra, Could you review the linting problems before merge the PR?

PR #2479, opened 2026-05-25
Add an env to configure sourcify URLs to support self hosted instances

cen1:
If you run a self hosted Sourcify instance it is currently not possible to use it as an ABI source. This makes the URL configurable instead of hardcoded.


falvaradorodriguez:
Thank you very much for the PR. It makes perfect sense. 

We’ve made some changes in the Sourcify Client (https://github.com/safe-global/safe-eth-py/pull/2522) and have already incorporated them. 

For this reason, I'm closing the PR as it is no longer necessary.

PR #2484, opened 2026-05-26
fix: Verify commits from Github action

luarx:
## Summary
Push verified commits from Github action "process issue


falvaradorodriguez (changes requested):
This will not solve the problem. We need to be able to commit from  `- name: Run pre-commit`

falvaradorodriguez (approved):

PR #2500, opened 2026-05-29
fix: consolidate address scripts and create PR after linter fixes

falvaradorodriguez:


.github/workflows/create_pr_with_new_address.yml
```diff
@@ -168,7 +166,17 @@ jobs:
...
-        uv run python .github/scripts/github_adding_addresses/commit_linter_fixes.py
+        uv run python .github/scripts/github_adding_addresses/create_pr_with_new_address.py --action commit-linter-fixes
+
+    - name: Create PR
+      if: steps.check-comment.outputs.result == 'true'
+      env:
+        GITHUB_PAT: ${{ secrets.TOKEN_GITHUB_SAFE_ETH_PY }}
+        GITHUB_REPOSITORY_NAME: ${{ github.repository }}
+        ISSUE_NUMBER: ${{ github.event.issue.number }}
+        ISSUE_BODY_INFO: ${{ steps.get-issue-inputs.outputs.result }}
```
moisses89, line 177: Could we validate the inputs from the issue? like we did for version.

falvaradorodriguez: Inputs are validated before being processed by the script. This input is the result of the previous process. There was no user interaction. What do you suggest we validate?

moisses89: > Inputs are validated before being processed by the script. 

so if somebody edit the issue, would be validated again, right?

falvaradorodriguez: > so if somebody edit the issue, would be validated again, right?

If someone edits the issues, the validation process (`validate_new_address_issue_input_data`) runs again. This process (`create_pr_with_new_address`) is only triggered by the ‘/execute’ command and uses the issue's value at the time of execution.

moisses89 (approved):

PR #2511, opened 2026-06-01
fix: safe-eth-py bugfixes batch 1 (caches, django, checksum addressing)

Uxio0:
## Summary

Independent, low-risk bugfixes and optimizations found during the `safe-eth-py` package review (batch 1).

## Changes

- **`eth/utils.py`** — bound the checksum-address LRU cache default (`1_000_000_000` → `500_000`, still env-configurable) to avoid unbounded growth / OOM in long-running indexers.
- **`safe/account_abstraction/safe_operation.py`** — drop `@cache` on the `get_safe_operation_hash` instance method (it pinned every `SafeOperation` via `self` in the cache key — a leak).
- **`eth/django/{models,validators}.py`** — catch `(TypeError, ValueError)` so `None`/wrong-type input raises a clean `ValidationError` instead of an uncaught `TypeError`; fix an uninterpolated `%s` in the error message.
- **`safe/signatures.py`** — `get_signing_address` uses `fast_to_checksum_address` instead of the slow eth-keys path.
- **`eth/multicall.py`** — validate/normalize addresses via `fast_to_checksum_address` instead of no-op `ChecksumAddress(HexAddress(HexStr(...)))` casts.
- **`eth/contracts/__init__.py`** — `multi_send_call_only_V1_5_0` now points at `MultiSendCallOnly_V1_5_0.json` (was the 1.4.1 file → wrong bytecode).
- **`safe/api/transaction_service_api.py`** — `post_transaction` is now using `0x` as expected

Related to PLA-1568

### DIFF
--- a/safe_eth/safe/account_abstraction/safe_operation.py
+++ b/safe_eth/safe/account_abstraction/safe_operation.py
@@ -1,7 +1,7 @@
 import dataclasses
 import datetime
 import logging
-from functools import cache, cached_property
+from functools import cached_property
 from typing import Optional, Union
 from zoneinfo import ZoneInfo
 
@@ -153,7 +153,6 @@ class SafeOperation:
             )
         return _domain_separator_cache[key]
 
-    @cache
     def get_safe_operation_hash(
         self, chain_id: int, module_address: ChecksumAddress
     ) -> bytes:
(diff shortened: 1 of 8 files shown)


safe_eth/safe/account_abstraction/safe_operation.py
```diff
@@ -153,7 +153,6 @@ def get_domain_separator(
             )
         return _domain_separator_cache[key]
 
-    @cache
     def get_safe_operation_hash(
```
moisses89, line 156: Why not a cached_property?

moisses89 (approved):
Just a minor comment, to ensure that the best solution is remove the cache totally.

safe_eth/safe/account_abstraction/safe_operation.py
Uxio0: Properties don't have arguments, that's why it cannot be used here

PR #2522, opened 2026-06-03
fix: migrate SourcifyClient to API v2

falvaradorodriguez:
Sourcify is running scheduled brownouts on API v1, causing `get_contract_metadata` and `async_get_contract_metadata` to silently return `None`. The old client hit `repo.sourcify.dev/contracts/{match_type}/...` which redirects to a v1 path that returns HTTP 503 during brownout windows:
  
  > "API v1 is temporarily unavailable during a scheduled brownout period.
  >  Please migrate to API v2. Full API docs: https://sourcify.dev/server/api-docs/swagger.json"
  
## Solution
  

 GET /server/v2/contract/{chainId}/{address}?fields=abi,metadata,compilation.name

  - `match == "exact_match"` → full match (`partial_match=False`)
  - `match == "match"` → partial match (`partial_match=True`)
  - Contract name resolved from `compilation.name`, with fallback to `metadata.settings.compilationTarget`
  - Removed `base_url_repo` parameter (no longer needed)

### DIFF
--- a/safe_eth/eth/clients/sourcify_client.py
+++ b/safe_eth/eth/clients/sourcify_client.py
@@ -1,6 +1,6 @@
 import os
 from functools import cache
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, Optional
 from urllib.parse import urljoin
 
 import aiohttp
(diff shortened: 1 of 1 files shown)


Uxio0 (dismissed):
@falvaradorodriguez it looks good, can you check if adding the url by environment variable for self hosted instances makes sense? https://github.com/safe-global/safe-eth-py/pull/2479/changes

If that's so it should be added in this PR too

Uxio0 (approved):

PR #2523, opened 2026-06-05
feat: add AsyncEthereumClient mirroring EthereumClient

Uxio0:
## Summary
Adds an asyncio-based `AsyncEthereumClient` that mirrors `EthereumClient`'s feature set using `AsyncWeb3`/`AsyncHTTPProvider` + `aiohttp`, while refactoring `EthereumClient` so the two share all I/O-free logic instead of duplicating it. Also fixes three bugs found during the refactor.

Closes PLA-1596.

## Async client
- `AsyncEthereumClient(EthereumClient)` and async managers subclass the sync ones, overriding **only** the I/O with `async_`-prefixed coroutines; every pure helper (payload building, response decoding, tx error mapping, trace selectors, manager builders/formatters) is reused, so no business logic is duplicated.
- `AsyncMulticall(Multicall)` overrides only the on-chain `eth_call`, so `async_batch_call` / `async_batch_call_same_function` keep **full parity** with the sync client: Multicall when available, JSON-RPC batch fallback, `force_batch_call` honored, and identical results — including raw revert bytes for failed calls.
- Per-event-loop `aiohttp` session with `aclose()` / `async with` lifecycle; POA middleware mirrored. Async tracing uses raw JSON-RPC (`AsyncWeb3` has no `.tracing`).
- Exported from `safe_eth.eth` alongside `get_auto_async_ethereum_client`.

## Refactor (EthereumClient / Multicall)
Extracted shared, I/O-free helpers used by both sync and async clients: JSON-RPC payload builders, batch-response processing/validation, `eth_call` decoding, tx error mapping, trace selectors, ERC20/721 builders & formatters, and Multicall result decoding.

## Bug fixes
- `get_total_transfer_history` double-counted a transfer matching both the `from` and `to` filters — now deduplicated on `(transactionHash, logIndex)`.
- `send_unsigned_transaction` returned an empty `HexBytes("")` (looked like success) when retries were exhausted — now raises.
- `functools.cache` on instance methods (`get_chain_id`, `get_client_version`, `is_eip1559_supported`, `get_singleton_factory_address`) pinned every client instance for the process lifetime — replaced with a per-instance cache + `clear_cache()`.

## Tests
- The existing sync suites (`TestERC20Module`, `TestTracingManager`, `TestEthereumClient`) are reused against the async client via a small `SyncCallProxy`, so the async I/O paths are exercised by the same assertions; only the handful of tests that mock synchronous internals are overridden to target the async equivalent.
- Added a sync-vs-async `batch_call` parity test (including the revert case).
- Full suite green against ganache: `ethereum_client` (sync), `async_ethereum_client`, `multicall`, and the Safe test base setUp — 114 passed.

## Notes
- `requests.Session` thread-safety, `get_balances` native-balance N+1, and an async `get_transfer_history` are tracked as follow-ups (PLA-1598, PLA-1599, PLA-1600).


safe_eth/eth/async_ethereum_client.py
```diff
@@ -0,0 +1,1121 @@
...
+            session = aiohttp.ClientSession()
+            self._async_http_sessions[loop] = session
+        return session
+
+    async def aclose(self) -> None:
+        """Close every open ``aiohttp`` session and the async providers."""
+        for session in list(self._async_http_sessions.values()):
+            if not session.closed:
+                await session.close()
+        self._async_http_sessions.clear()
```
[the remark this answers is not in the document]
Uxio0: Agree, done in 1197f3e

safe_eth/eth/ethereum_client.py
```diff
@@ -1341,62 +1555,33 @@ def raw_batch_request(
...
     def current_block_number(self):
         return self.w3.eth.block_number
 
-    @cache
     def get_chain_id(self) -> int:
         """
         :return: ChainId returned by the RPC `eth_chainId` method. It should never change, so it's cached.
         """
-        return int(self.w3.eth.chain_id)
+        if "chain_id" not in self._cache:
```
falvaradorodriguez, line 1574: functools.cache was thread-safe, with this _cache dict we lose that. Do you see any problem here?

safe_eth/eth/async_ethereum_client.py
```diff
@@ -0,0 +1,1139 @@
...
+        count: Optional[int] = None,
+    ) -> List[FilterTrace]:
+        """
+        Async counterpart of :meth:`TracingManager.trace_filter`.
+
+        .. note::
+            Tracing is issued as a raw JSON-RPC call (``AsyncWeb3`` has no
+            ``tracing`` module). A node error (e.g. ``trace_filter`` unsupported)
+            therefore surfaces as a ``ValueError`` from the batch layer, not as
+            web3's ``Web3RPCError`` raised by the sync path.
```
falvaradorodriguez: Just to check, could this cause a problem for our lib users?

safe_eth/eth/ethereum_client.py
```diff
@@ -2020,7 +2210,11 @@ def send_unsigned_transaction(
                     address, block_identifier=block_identifier
                 )
                 number_errors -= 1
-        return HexBytes("")
+        # Retries exhausted without a successful send. Raise instead of returning an
+        # empty hash that would look like a success to callers.
+        raise ValueError(
```
falvaradorodriguez, line 2215: Should we indicate this as a breaking change?

falvaradorodriguez (dismissed):

safe_eth/eth/ethereum_client.py
Uxio0: functools cache was as safe as this, as it does not guarantee the wrapped function runs only once. I don't see a problem

safe_eth/eth/ethereum_client.py
Uxio0: The old return HexBytes("") was unreachable in normal flow, so it's more a bugfix than a behaviour change

safe_eth/eth/async_ethereum_client.py
Uxio0: Only for the ones porting to async, so I'm documenting it on the docstring: 5e027f4

falvaradorodriguez (approved):

PR #2576, opened 2026-07-10
Allow building AsyncEthereumClient without blocking network calls

Uxio0:
## Motivation

`AsyncEthereumClient.__init__` performed a synchronous `eth_chainId` call to decide whether to inject `ExtraDataToPOAMiddleware` (skipped on Mainnet). Building the client inside a running event loop (e.g. a FastAPI handler) blocked it, up to the provider timeout if the RPC was down. safe-queue-service builds clients lazily per request with RPC urls from the Config Service, so this blocked the loop in practice.

## Changes

- Always inject `ExtraDataToPOAMiddleware`, on every network, so no network detection is needed on `__init__`. Constructing `EthereumClient`/`AsyncEthereumClient` now performs zero network I/O. Middleware setup is shared in `EthereumClient._adjust_middlewares`.
- Add `AsyncMulticall.async_create` and `AsyncEthereumClient.async_get_multicall`: first use of Multicall on the async client (via `async_batch_call`/`async_batch_call_same_function`) detects the Multicall address with the async RPC methods instead of blocking the event loop with the sync ones. Sync `multicall` property and `async_get_multicall` share the same memoized instance.
- `Multicall.__init__` no longer calls `get_network()` when an explicit address is provided.
- Tests: construction with an unreachable RPC url performs no requests (sync and async providers are patched to fail on any call), works inside a running event loop, and `async_get_multicall` never uses the blocking sync methods.

## Behaviour change on Mainnet

`ExtraDataToPOAMiddleware` unconditionally renames `extraData` to `proofOfAuthorityData` in `eth_getBlockByHash`/`eth_getBlockByNumber`/`eth_subscribe` responses. Mainnet block dicts from `w3.eth.get_bl

### DIFF
--- a/safe_eth/eth/async_ethereum_client.py
+++ b/safe_eth/eth/async_ethereum_client.py
@@ -49,7 +49,6 @@ from web3.exceptions import (
     TransactionNotFound,
     Web3Exception,
 )
-from web3.middleware import ExtraDataToPOAMiddleware
 from web3.types import (
     BlockData,
     BlockIdentifier,
@@ -500,6 +499,10 @@ class AsyncEthereumClient(EthereumClient):
     The ``aiohttp`` session used for JSON-RPC batches is created lazily, one per
     running event loop, so the same instance works across loops (and is safe to
     create outside a running loop, e.g. in ``__init__``).
+
+    Constructing the client performs no network I/O, so it can be built inside a
+    running event loop (e.g. in a FastAPI handler) without blocking it, even if
+    the RPC url is not reachable.
     """
 
     # Narrow the manager types overridden in __init__ for the type checker
@@ -517,7 +520,7 @@ class AsyncEthereumClient(EthereumClient):
         use_request_caching: bool = True,
         batch_request_max_size: int = 500,
     ):
-        # Builds the blocking w3/slow_w3 + detects the network (cached chainId)
+        # Builds the blocking w3/slow_w3
         super().__init__(
             ethereum_node_url,
             provider_timeout=provider_timeout,
@@ -545,18 +548,7 @@ class AsyncEthereumClient(EthereumClient):
         self.async_w3: AsyncWeb3 = AsyncWeb3(self.async_w3_provider)
         self.async_slow_w3: AsyncWeb3 = AsyncWeb3(self.async_w3_slow_provider)
 
-        for async_w3 in self.async_w3, self.async_slow_w3:
-            # Don't spend resources converting dictionaries to attribute dictionaries
-            async_w3.middleware_onion.remove("attrdict")
-
-        # POA middleware (reuse the network already detected by the sync __init__)
-        try:
-            inject_poa = self.get_network() != EthereumNetwork.MAINNET
-        except (IOError, OSError):
-            inject_poa = True
-        if inject_poa:
-            for async_w3 in self.async_w3, self.async_slow_w3:
-                async_w3.middleware_onion.inject(ExtraDataToPOAMiddleware, layer=0)
+        self._adjust_middlewares(self.async_w3, self.async_slow_w3)
 
         # Replace the sync managers built by super().__init__ with async-capable ones
         self.erc20 = AsyncErc20Manager(self)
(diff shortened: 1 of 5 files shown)


safe_eth/eth/ethereum_client.py
```diff
@@ -1497,28 +1500,31 @@ def __init__(
...
+          attribute dictionaries.
+        - Inject ``ExtraDataToPOAMiddleware``: required for PoA-based chains like
+          Polygon or BNB Chain. It's always injected to avoid a blocking
+          ``eth_chainId`` call during ``__init__`` just to detect the network.
+          On Mainnet its only effect is that block responses expose
+          ``proofOfAuthorityData`` instead of ``extraData``.
+        """
+        for w3 in w3s:
+            w3.middleware_onion.remove("attrdict")
+            w3.middleware_onion.inject(ExtraDataToPOAMiddleware, layer=0)
```
falvaradorodriguez, line 1526: Previously, this was only injected if self.get_network() != EthereumNetwork.MAINNET. Is that a problem?

safe_eth/eth/ethereum_client.py
```diff
@@ -1497,28 +1500,31 @@ def __init__(
...
-        # It may also be needed for other EVM compatible blockchains like Polygon or BNB Chain (Binance Smart Chain).
-        try:
-            if self.get_network() != EthereumNetwork.MAINNET:
-                for w3 in self.w3, self.slow_w3:
-                    w3.middleware_onion.inject(ExtraDataToPOAMiddleware, layer=0)
-        except (IOError, OSError):
-            # For tests using dummy connections (like IPC)
-            for w3 in self.w3, self.slow_w3:
-                w3.middleware_onion.inject(ExtraDataToPOAMiddleware, layer=0)
+        self._adjust_middlewares(self.w3, self.slow_w3)
```
falvaradorodriguez, line 1503: We are implicitly not caching the chain ID, since we are not calling the:
```
def get_network(self) -> EthereumNetwork:                                                                                                                                                                                                                                                                                 
      return EthereumNetwork(self.get_chain_id())                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                            
      if "chain_id" not in self._cache:                                                                                                                                                                                                                                                                                     
          self._cache["chain_id"] = int(self.w3.eth.chain_id)  # bloqueante                                                                                                                                                                                                                                                 
      return self._cache["chain_id"]   
```
 I understand that this isn’t a problem, but I’d like to mention it here all the same.

safe_eth/eth/async_ethereum_client.py
```diff
@@ -580,6 +577,25 @@ def multicall(self) -> Optional["AsyncMulticall"]:  # type: ignore # noqa F821
...
+        Async counterpart of :attr:`multicall`: detect the Multicall address with
+        the async RPC methods, so a running event loop is never blocked. Memoized
+        in the same slot used by the cached property, so both share one instance.
+        """
+        if "multicall" not in self.__dict__:
+            from .multicall import AsyncMulticall
+
+            try:
+                multicall: Optional["AsyncMulticall"] = (
+                    await AsyncMulticall.async_create(self)
```
falvaradorodriguez, line 591: Is an `asyncio.Lock` useful to prevent two simultaneous creations?

safe_eth/eth/ethereum_client.py
Uxio0: No, it shouldn't be as documented in the class. Tested it with multiple mainnet RPCs and it's working properly

safe_eth/eth/ethereum_client.py
Uxio0: It shouldn't be a problem

safe_eth/eth/async_ethereum_client.py
Uxio0: I think benefit is marginal in this case. Creating an AsyncMulticall is quite cheap, so worst case scenario we create multiple instances but just store one

falvaradorodriguez (approved):

PR #2577, opened 2026-07-10
Use Safe as eth_call sender for EIP-1271 contract signature validation

johanneskares:
## Motivation

Fixes the root cause behind safe-global/safe-transaction-service#2937.

On-chain, `checkSignatures` has the **Safe itself** call `isValidSignature` on its owner contracts, so `msg.sender` is always the Safe. The off-chain check in `SafeSignatureContract._check_eip1271` performs a from-less `eth_call` (`msg.sender = address(0)`), which diverges from the on-chain semantics it is meant to predict.

For most owner contracts this makes no difference, but signatures from **`msg.sender`-dependent owners always fail validation**. The canonical example is the `SafeWebAuthnSharedSigner` from safe-modules (the standard gas-optimized ERC-4337 passkey setup): it stores each Safe's P-256 public key in the *Safe's* storage and reads it back via `ISafe(msg.sender).getStorageAt(...)` — in a from-less call there is no configuration to find, so pure-passkey Safes cannot propose transactions or register delegates with the Transaction Service at all.

## Change

`SafeSignatureContract.is_valid()` (sync and async) already receives `safe_address` — other signature types like `SafeSignatureApprovedHash` use it — but the contract-signature path ignored it. This PR threads it into the EIP-1271 `eth_call` as the sender when provided:

- `safe_address` given → `isValidSignature` is called with `from = safe_address`, mirroring on-chain `checkSignatures`.
- `safe_address` not given → unchanged from-less call.

Owner contracts that ignore `msg.sender` (e.g. a Safe validating through its fallback handler) are unaffected.

## Tests

Added sync + async tests deploying a minimal EIP-1271 signer that returns the magic value only when `msg.sender` matches a configured address (emulating the shared signer's behavior), asserting that validation fails without `safe_address` and succeeds with it. All existing tests in `test_safe_signature.py` pass (38 passed locally against a local node).

## Notes for safe-transaction-service

The service's proposal endpoints already pass `safe_address` to `is_valid`, so they gain shared-signer support with a dependency bump and no code change. The delegates v2 endpoint currently passes the delegator as second argument (`is_valid(ethereum_client, owner)`); supporting shared-signer delegators there needs a small follow-up passing the request's `safe` — happy to submit that PR as well.

### DIFF
--- a/safe_eth/safe/safe_signature.py
+++ b/safe_eth/safe/safe_signature.py
@@ -578,6 +578,7 @@ class SafeSignatureContract(SafeSignatureContractMixin, SafeSignature):
         function_signature: str,
         data: bytes,
         signature: bytes,
+        safe_address: Optional[str] = None,
     ) -> bool:
         """
         Attempt to validate an EIP-1271 signature using a specific CompatibilityFallbackHandler and function signature.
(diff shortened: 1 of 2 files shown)


johanneskares:

johanneskares:

johanneskares:
recheck

Issue #2139, opened 2025-11-12
[New chain]: hyperEVM 1.3.0 canonical


Fbartoli:
### Summary

Adding hyperevm 1.3.0 canonical

### Chain ID

999

### Chain detail URL

https://chainlist.org/chain/999

### RPC URL

https://hyperliquid.drpc.org

### Version

1.3.0

### Address (Master copy)

0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552

### Deployment Tx hash (Master copy)

0x8ca524ed5be0fbd74d56001aac9e7f389ed0f33341e77227232eb0c7e7a9f253

### Block explorer URL (Master copy)

https://www.hyperscan.com/tx/0x8ca524ed5be0fbd74d56001aac9e7f389ed0f33341e77227232eb0c7e7a9f253

### Address (Master copy L2)

0x3E5c63644E683549055b9Be8653de26E0B4CD36E

### Deployment Tx hash (Master copy L2)

0x2845931ddf596e21d09830d8c1a3257b60a9b3296f65753e7c2ac9cfad834541

### Block explorer URL (Master copy L2)

https://www.hyperscan.com/tx/0x2845931ddf596e21d09830d8c1a3257b60a9b3296f65753e7c2ac9cfad834541

### Address (Proxy factory)

0xa6B71E26C5e0845f74c812102Ca7114b6a896AB2

### Deployment Tx hash (Proxy factory)

0x69c2fa1b9df78aa0dd500ca784ec081f5062d86f63a76805f4eb081662467c04

### Block explorer URL (Proxy factory)

https://hyperevmscan.io/tx/0x69c2fa1b9df78aa0dd500ca784ec081f5062d86f63a76805f4eb081662467c04

PR #2595, opened 2026-07-25
fix: Add missing HyperEVM entry to transaction service network map

Primata:
HyperEVM (chain 999) is supported everywhere else in the library — it is
in the `EthereumNetwork` enum, and its 1.3.0/1.4.1 Safe deployments were
added in #2139-#2144 — but it was never added to
`TransactionServiceApi.NETWORK_SHORTNAME`.

Because `_get_url_by_network` resolves the base URL from that map,
`TransactionServiceApi(EthereumNetwork.HYPEREVM)` raises
`EthereumNetworkNotSupported`, which in turn makes `safe-cli` refuse to
enter tx-service mode on HyperEVM.

The endpoint exists and is live: safe-config lists chain 999 with
`transactionService: https://api.safe.global/tx-service/hyper`, and
`GET /tx-service/hyper/api/v1/about/` returns 200.


Primata:

Primata:
recheck

falvaradorodriguez (approved):

Uxio0 (approved):

PR #2715, opened 2026-08-25
Add Tempo transaction service network slugs

0xalpharush:
## Summary

Safe Transaction Service endpoints are live for both Tempo networks, but the client-side network map did not include their service slugs. As a result, constructing `TransactionServiceApi` for either network raised `EthereumNetworkNotSupported`.

This change adds:

- `EthereumNetwork.TEMPO_MAINNET` -> `tempo`
- `EthereumNetwork.TEMPO_TESTNET_MODERATO` -> `tempo-moderato`

The resulting base URLs are:

- `https://api.safe.global/tx-service/tempo/`
- `https://api.safe.global/tx-service/tempo-moderato/`

## Testing

- Added constructor coverage for both Tempo service URLs.
- Python syntax compilation passed.
- `git diff --check` passed.
- Full pytest could not run because the environment cannot download unavailable dependencies.


0xalpharush:

PR #2726, opened 2026-09-02
Validate contract signatures against the verifying Safe's EIP-1271 convention

Uxio0:
# Validate contract signatures against the verifying Safe's EIP-1271 convention

## Problem

`SafeSignatureContract.is_valid` tries both EIP-1271 entrypoints and returns `True` if **either** answers:

1. `isValidSignature(bytes32,bytes)` with `safe_hash`
2. `isValidSignature(bytes,bytes)` with `safe_hash_preimage`

On-chain only one is ever called, and the version of the Safe **doing the verifying** decides which:

- **Below 1.5.0** (1.3.0, 1.4.x) — `checkNSignatures` calls the legacy `isValidSignature(data, contractSignature)` with the whole preimage (`Safe.sol:315` in v1.4.1, `GnosisSafe.sol:285` in v1.3.0 — identical branches, the only v1.4.1 addition is the `GS027` preimage assertion).
- **From 1.5.0** — `checkContractSignature` calls `isValidSignature(bytes32 dataHash, ...)` only (`Safe.sol:270`), and its `CompatibilityFallbackHandler` no longer declares the legacy overload at all.

So a signature can be "valid" here and still revert `GS024` on the Safe that has to verify it. Consumers that store it end up below threshold with nothing reported at write time.

## What this changes

**`safe_version` on `is_valid`.** Only the entrypoint that version calls on-chain is checked. It goes on the abstract `SafeSignature` / `SafeSignatureAsync` and on all ten implementations, the same way `safe_address` already works: validation-time context that only one signature type reads and the rest ignore.

**`safe_version=None` keeps the current behaviour.** Existing callers are unaffected — the library cannot always know the verifying Safe's version, so the permissive default stays. Callers that do know should pass it.

**`uses_bytes32_eip1271()`** is the single home for the version rule and for its failure mode. `VERSION()` is an arbitrary on-chain string, so it raises `CannotCheckEIP1271ContractSignature` (an existing, until now unused, `SafeSignatureException` subclass) instead of leaking `packaging`'s `InvalidVersion` and making every consumer reimplement the same translation.

## The cross-version case comes out for free

A verifying Safe below 1.5.0 with an owner Safe on 1.5.0 can never work: the verifying Safe calls the legacy overload, which a 1.5.0 signer does not declare. Today `is_valid` accepts it through attempt 1, which is a permanent false accept. Keying only off the verifying Safe's version rejects it with no extra logic, because we call the entrypoint that will actually be used and it reverts.

| Verifying Safe | Owner Safe | Called on-chain | Works | Before | After |
| --- | --- | --- | --- | --- | --- |
| pre-1.5.0 | pre-1.5.0 | `(bytes preimage, sig)` | yes | accepts preimage and hash | preimage only |
| pre-1.5.0 | 1.5.0+ | `(bytes preimage, sig)`, overload missing | never | **accepts** | rejected |
| 1.5.0+ | pre-1.5.0 | `(bytes32 hash, sig)` | yes | accepts hash and preimage | hash only |
| 1.5.0+ | 1.5.0+ | `(bytes32 hash, sig)` | yes | accepts hash only | hash only |

## Tests

`TestSafeContractSignature` and `TestSafeContractSignatureAsync` cover v1.3.0, v1.4.1 and v1.5.0 against the deployed contracts:

- the version-correct nesting is accepted, the other is rejected
- `safe_version=None` still accepts both on a pre-1.5.0 signer (back-compat guard)
- a 1.5.0 owner Safe is rejected for a 1.4.1 verifier, and the test pins that without a version it is still accepted, which is the false accept above

`277 passed, 12 skipped` in `safe_eth/safe/tests/`.

## Notes for the reviewer

- No production code in this repo calls `is_valid`, so the change is additive to the public API.
- The `mypy` pre-commit hook is `language: system` and fails locally with `Executable mypy not found`. Ran separately with the hook's flags: both changed files clean. The one error reported, `safe_eth/eth/tests/utils.py:56`, is pre-existing in a file this PR does not touch.

Consumed by safe-global/safe-queue-service PLA-1907, which needs a release of this.

### DIFF
--- a/safe_eth/safe/safe_signature.py
+++ b/safe_eth/safe/safe_signature.py
@@ -621,6 +621,7 @@ class SafeSignatureContract(SafeSignatureContractMixin, SafeSignature):
         function_signature: str,
         data: bytes,
         signature: bytes,
+        expected_magic_value: HexBytes,
     ) -> bool:
         """
         Attempt to validate an EIP-1271 signature using a specific CompatibilityFallbackHandler and function signature.
@@ -630,6 +631,11 @@ class SafeSignatureContract(SafeSignatureContractMixin, SafeSignature):
         :param function_signature: The ABI function signature to call.
         :param data: The data or hash to be validated, depending on the Safe version.
         :param signature: The contract signature payload to validate.
+        :param expected_magic_value: The value the verifying Safe requires back for this
+            entrypoint. Each interface has its own: `ISignatureValidator` declares
+            `0x20c13b0b` up to 1.4.1 and `0x1626ba7e` from 1.5.0, and the Safe compares
+            against that one exactly. Accepting either would pass a signer that returns the
+            other, which then reverts `GS024` on the Safe.
         :return: True on successful validation; otherwise False.
         """
         fallback_handler = fallback_handler_getter(ethereum_client.w3, self.owner)
@@ -648,10 +654,7 @@ class SafeSignatureContract(SafeSignatureContractMixin, SafeSignature):
             )
             return False
 
-        return result in (
-            self.EIP1271_MAGIC_VALUE,
-            self.EIP1271_MAGIC_VALUE_UPDATED,
-        )
+        return HexBytes(result) == expected_magic_value
 
     def is_valid(
         self,
@@ -691,6 +694,7 @@ class SafeSignatureContract(SafeSignatureContractMixin, SafeSignature):
             "isValidSignature(bytes32,bytes)",
             bytes(self.safe_hash),
             bytes(self.contract_signature),
+            self.EIP1271_MAGIC_VALUE_UPDATED,
         ):
             return True
 
@@ -702,6 +706,7 @@ class SafeSignatureContract(SafeSignatureContractMixin, SafeSignature):
             "isValidSignature(bytes,bytes)",
             bytes(self.safe_hash_preimage),
             bytes(self.contract_signature),
+            self.EIP1271_MAGIC_VALUE,
         ):
             return True
 
@@ -781,6 +786,7 @@ class SafeSignatureContractAsync(SafeSignatureContractMixin, SafeSignatureAsync)
         function_signature: str,
         data: bytes,
         signature: bytes,
+        expected_magic_value: HexBytes,
     ) -> bool:
         """
         Attempt to validate an EIP-1271 signature using a specific CompatibilityFallbackHandler and function signature.
@@ -790,6 +796,11 @@ class SafeSignatureContractAsync(SafeSignatureContractMixin, SafeSignatureAsync)
         :param function_signature: The ABI function signature to call
         :param data: The data or hash to be validated, depending on the Safe version.
         :param signature: The contract signature payload to validate.
+        :param expected_magic_value: The value the verifying Safe requires back for this
+            entrypoint. Each interface has its own: `ISignatureValidator` declares
+            `0x20c13b0b` up to 1.4.1 and `0x1626ba7e` from 1.5.0, and the Safe compares
+            against that one exactly. Accepting either would pass a signer that returns the
+            other, which then reverts `GS024` on the Safe.
         :return: True on successful validation; otherwise False.
         """
         fallback_handler = fallback_handler_getter(web3, self.owner)
@@ -808,10 +819,7 @@ class SafeSignatureContractAsync(SafeSignatureContractMixin, SafeSignatureAsync)
             )
             return False
 
-        return result in (
-            self.EIP1271_MAGIC_VALUE,
-            self.EIP1271_MAGIC_VALUE_UPDATED,
-        )
+        return HexBytes(result) == expected_magic_value
 
     async def is_valid(
         self,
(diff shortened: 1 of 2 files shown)


safe_eth/safe/safe_signature.py
```diff
@@ -614,40 +657,53 @@ def is_valid(
...
+            try_bytes32 = uses_bytes32_eip1271(safe_version)
+            try_legacy = not try_bytes32
+
+        if try_bytes32 and self._check_eip1271(
+            ethereum_client,
+            get_compatibility_fallback_handler_contract,
+            "isValidSignature(bytes32,bytes)",
+            bytes(self.safe_hash),
+            bytes(self.contract_signature),
         ):
```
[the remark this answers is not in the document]
Uxio0: Agree, fixed on f85e4eb762e804845bc9a9ad809324632fb687f1

Uxio0:
Good catch, and it holds. Checked the constants in `safe-smart-account`:

| Verifying Safe | Calls | Requires |
| --- | --- | --- |
| v1.3.0 / v1.4.1 | `isValidSignature(bytes,bytes)` | `0x20c13b0b` |
| v1.5.0 | `isValidSignature(bytes32,bytes)` | `0x1626ba7e` |

Both are named `ISignatureValidator.EIP1271_MAGIC_VALUE`, but they are different values per version (`ISignatureValidator.sol:6` in v1.3.0/v1.4.1, `:7` in v1.5.0), and the Safe compares against its own one exactly — `Safe.sol:315` in v1.4.1, `Safe.sol:270` in v1.5.0.

Worth noting no Safe signer can hit this: the v1.4.1 handler returns `UPDATED_MAGIC_VALUE` (`0x1626ba7e`) from the `bytes32` entrypoint and `EIP1271_MAGIC_VALUE` from the legacy one, and the v1.5.0 handler returns `0x1626ba7e`. So the pairing always matches for a Safe owner. But an owner can be any EIP-1271 contract, and one answering the other interface's value was reported valid here and then reverted `GS024` on-chain — the same false accept this PR is about, one layer down.

Fixed in f85e4eb: `_check_eip1271` takes the expected value and compares against it alone, sync and async.

On testing it: the existing suite already covers the pairing being right, since a crossed pairing would fail every contract-signature test at 1.4.1 or 1.5.0. Reaching the rejection needs a signer no Safe can be, so there is a stubbed fallback handler returning the legacy value from the `bytes32` entrypoint, asserted valid before the stub and invalid under it.

safe_eth/safe/safe_signature.py
```diff
@@ -605,49 +654,61 @@ def _check_eip1271(
...
-                return True
+            return True
+
+        # Dropped from the 1.5.0 `CompatibilityFallbackHandler`, so a 1.5.0 signer does not
+        # declare this overload at all and can never answer here
+        if try_legacy and self._check_eip1271(
+            ethereum_client,
+            get_compatibility_fallback_handler_V1_4_1_contract,
+            "isValidSignature(bytes,bytes)",
+            bytes(self.safe_hash_preimage),
```
falvaradorodriguez: Nit: parse_signature defaults safe_hash_preimage to safe_hash, so a caller using the two arg form with a pre 1.5.0 safe_version gets False for a valid signature. WDYT?

falvaradorodriguez (dismissed):

safe_eth/safe/safe_signature.py
Uxio0: Good catch, and it is worse than a nit: for a caller using the two arg form, `safe_hash_preimage` came back as `safe_hash`, which looks like a real preimage but is not one. With a `safe_version` below 1.5.0 only the legacy `isValidSignature(bytes,bytes)` is checked, and that entrypoint takes the whole preimage — the signer hashes whatever it gets, so a valid signature was reported as invalid.

Fixed in 317dd4b:

- `parse_signature` no longer defaults the preimage to the hash, so "not given" stops looking like "the preimage is the hash". `safe_hash_preimage` is now `Optional[HexBytes]`.
- New `_legacy_eip1271_data()` on the mixin is the only way into the legacy entrypoint, and raises `CannotCheckEIP1271ContractSignature` when the preimage is unknown. No more silent `False`.
- With `safe_version=None` the legacy attempt is skipped when the preimage is unknown, so the `bytes32` entrypoint still answers and the "try both" behaviour is unchanged.
- Four new tests, sync and async: the raise at 1.4.1 without the preimage, `False` with no version, valid once the preimage is passed, and 1.5.0 not needing it at all.

The four `test_*_deprecated` were relying on the old default. Their signer is a Safe v1.1.1, which only declares the legacy entrypoint, and they `signMessage` `safe_tx_hash` itself — so the hash really is the preimage there and they now pass it explicitly.

Nothing to change downstream: `safe-transaction-service` and `safe-queue-service` always pass `safe_hash_preimage=`, and the two places that pass the hash as the preimage only call `export_signature()`, which does not use it.

falvaradorodriguez (approved):

PR #2753, opened 2026-09-22
fix: Wrap transport errors in the library's own exception types

Uxio0:
## What

The raw JSON-RPC call sites of `EthereumClient` and `AsyncEthereumClient` leaked `requests` and `aiohttp` exceptions. A caller had to import the HTTP library to handle a node being down, timing out or not resolving.

Six call sites are covered, sync and async:

| Method | Connection failure | `not response.ok` |
| --- | --- | --- |
| `batch_call_custom` | `EthereumClientConnectionException` | `EthereumClientConnectionException` (was the builtin `ConnectionError`) |
| `Erc20Manager.get_info` | `InvalidERC20Info` | `InvalidERC20Info` (unchanged) |
| `raw_batch_request` | `EthereumClientConnectionException` | `ValueError` (unchanged) |

Each site keeps the exception it already raised for a bad status, so nothing that works today stops working. The original exception is kept as the cause.

`wrap_http_exceptions` lives in `safe_eth/util/http.py`, next to `prepare_http_session`. It covers both transports and the JSON decoding of the response, which fails the same way when a proxy answers with HTML instead of JSON.

In `get_info` the wrapper also covers `_build_info_payload`, because building the payload calls `build_transaction`, which queries the node through web3.

## The new exception type

```python
class EthereumClientConnectionException(EthereumClientException, ConnectionError):
```

The builtin `ConnectionError` base keeps `batch_call_custom` compatible with callers that catch the exception it raised before. The `EthereumClientException` base puts it in the library hierarchy, and makes it a `ValueError` like the rest of it.

## Behavior changes worth knowing

- `batch_call_custom` no longer raises the builtin `ConnectionError`. `except ConnectionError` still catches the new type.
- A connection failure in `get_info` now raises `InvalidERC20Info` instead of a `requests` or `aiohttp` exception.
- `EthereumClientConnectionException` is a `ValueError`, so the broad `except ValueError` handlers inside the library catch it too. A node that is down during `erc721.get_info`, `Safe.retrieve_all_info` or the oracles now surfaces as `InvalidERC721Info`, `CannotRetrieveSafeInfoException` or `CannotGetPriceFromOracle` instead of propagating. Consumers that treat those as permanent, like `safe-transaction-service` writing a `TokenNotValid` row, will do it on a transient outage.
- `get_total_transfer_history` still raises `requests.exceptions.ReadTimeout`. It queries the node through the web3 provider, not through these call sites, and the docstring now says so. Covering that needs a provider level wrapper, which is a separate decision.

## Tests

New `TestTransportExceptions`, reused against the `aiohttp` paths in `test_async_ethereum_client.py` through `SyncCallProxy`. It points the client at a hostname that does not resolve, so the transport really fails and the failure is not mocked. Two extra tests mock the transport: one to return a non ok status, and one to stub the payload building so the batch request itself is reached.

Related: [PLA-2011](https://linear.app/safe-global/issue/PLA-2011/safe-eth-py-wrap-transport-errors-in-the-librarys-own-exception-type)


safe_eth/util/http.py
```diff
@@ -1,8 +1,38 @@
...
+    cannot be reached or its response cannot be decoded, so callers only deal with
+    exceptions from this library. The original exception is kept as the cause.
+
+    :param url: Url being requested, used for the error message
+    :param exception_class: Exception raised instead of the HTTP library one
+    """
+    try:
+        yield
+    except HTTP_TRANSPORT_EXCEPTIONS as exc:
+        raise exception_class(f"Error connecting to {url}: {exc!r}") from exc
```
moisses89: I think JSONDecodeError should not be included here, I mean it is not a connection error right?

Uxio0: `requests.exceptions.JSONDecodeError` comes from `RequestException` and it can be originated by incomplete JSON, for example. So as we don't expect incomplete or malformed JSON I think it's alright to treat them as transport errors

Uxio0: Good point on the wording anyway: the message said `Error connecting to`, which is wrong for a response that arrived and could not be decoded. Changed it to `Error querying` in 99ee850.

moisses89 (approved):

PR #2760, opened 2026-09-23
fix: Raise Etherscan client errors and stop flaky network tests

Uxio0:
## Why

`test_etherscan_get_contract_metadata` fails often in CI with `AttributeError: 'NoneType' object has no attribute 'name'`, and passes when it is run again. Three causes:

1. The Etherscan client only detected the `Max rate limit reached` message. Other rate limit messages (`Max calls per sec rate limit reached (5/sec)`, the daily limit), HTTP 429 and any other API error returned `None`. So the test got `None` instead of `EtherscanRateLimitError`, and the `skipTest` never ran.
2. `@pytest.mark.flaky(reruns=5, delay=2)`: `delay` is not a valid kwarg for pytest-rerunfailures, it is ignored. Reruns ran with no delay, inside the same rate limit window.
3. The 4 jobs of the CI matrix call Etherscan at the same time with the same API key.

## Changes

Etherscan client (`safe_eth/eth/clients/etherscan_client_v2.py`):
- Any message with `rate limit`, and HTTP 429, raise `EtherscanRateLimitError`
- Other non ok HTTP responses raise the new `EtherscanHttpError` (with `status_code`)
- Other API errors (e.g. `Invalid API Key`) raise `EtherscanClientException`
- `None` is only returned when the contract is not verified
- Transport errors are wrapped with `wrap_http_exceptions` as the new `EtherscanConnectionError` (also a builtin `ConnectionError`)
- `_retry_request` raises `EtherscanRateLimitError` after the last attempt, instead of returning `None`
- Same changes for `AsyncEtherscanClientV2`

Tests and CI:
- Fix the flaky marker kwarg: `reruns_delay`
- New `network` pytest marker for tests hitting external services: Etherscan, Blockscout, Sourcify, ENS, CowSwap, 4337 bundler, Transaction Service API, and the mainnet/polygon RPC node tests. CI runs them only on Python 3.13, the other jobs run `-m "not network"`
- The markers are set per matrix entry in `include:`. A `python-version == 3.13 && '' || 'not network'` expression does not work: GitHub Actions treats `''` as false, so it always returns `'not network'`
- Etherscan network tests skip on rate limit, connection and HTTP errors. An invalid API key still fails the CI
- Offline unit tests for response processing, HTTP errors, retries and connection errors (sync and async)

## Breaking change

Callers that relied on `None` for Etherscan errors now get an exception. All new exceptions subclass `EtherscanClientException`.

`safe-decoder-service` catches only `(OSError, EtherscanRateLimitError)` in its Etherscan -> Sourcify -> Blockscout fallback, so an HTTP error or invalid key would stop the fallback. It must be fixed before upgrading: PLA-2030.


safe_eth/eth/clients/etherscan_client_v2.py
```diff
@@ -57,30 +79,76 @@ def build_url(self, query: str) -> str:
...
+
+        # `result` holds the error message, but it can be null for some errors. `message`
+        # can be null too, so `or ""` is needed on top of the `.get` default
+        message = (
+            result
+            if isinstance(result, str)
+            else str(response_json.get("message") or "")
+        )
+        lowered_message = message.lower()
+        if "rate limit" in lowered_message:
```
falvaradorodriguez: Could the broader "rate limit" match end up retrying the daily limit too? "Max daily rate limit reached..." now raises EtherscanRateLimitError, so _retry_request would sleep 5 s twice before re-raising. Since a daily quota won't reset in 10 s, once it's exhausted every get_contract_* call with the default retry=True would block for ~10 s and still fail (on main that message returned None right away). Would it make sense to raise it without retrying, for example a EtherscanDailyRateLimitError(EtherscanRateLimitError) that _retry_request re-raises immediately, or to only retry the per-second limit and HTTP 429?

safe_eth/eth/clients/etherscan_client_v2.py
```diff
@@ -57,30 +79,76 @@ def build_url(self, query: str) -> str:
...
-            except EtherscanRateLimitError as exc:
-                if not retry:
-                    raise exc
-                else:
-                    time.sleep(5)
+            except EtherscanRateLimitError:
+                if not retry or attempt == last_attempt:
+                    raise
+                time.sleep(5)
         return None
```
falvaradorodriguez:
```suggestion
```

falvaradorodriguez:
There’s also an issue with the GitHub configuration. This is now pending: 
Screenshot 2026-09-24 at 10 49 26

falvaradorodriguez (approved):

safe_eth/eth/clients/etherscan_client_v2.py
Uxio0: Good catch. Added `EtherscanDailyRateLimitError(EtherscanRateLimitError)` in 3456077. `_retry_request` raises it with no retry. The per second limit and HTTP 429 are still retried.

safe_eth/eth/clients/etherscan_client_v2.py
Uxio0: mypy needs a return there (`Missing return statement`), so I changed the loop instead in 3456077: 2 attempts with retry inside the loop, and the last attempt outside, so its error is raised directly. No dead line now.

